PR #47, opened 2015-11-12
adding varscan pipeline

Jeltje:
contains 
`varscan.py` wrapper script that runs the jeltje/varscan docker container on encrypted bam inputs from S3. Outputs to local dir or encrypted S3
`launch_varscan_mesos.sh` to run on toil mesos cluster
`config.txt` example input S3 bam files


jvivian:
This PR is based on Toil v3.0.7 and should be based against that branch instead of Master.

Jeltje:
Ah. Will do!

PR #73, opened 2015-12-30
ADAM/GATK pipeline, first cut

fnothaft:
Don't expect this to work yet. Will ping @almussel @jvivian @hannes-ucsc @heuermh and @benedictpaten offline with more details.

This builds off of #72. The big thing I wanted to get out there was the package refactoring. I made toil-scripts itself a package, and then changed a bunch of file names to conform with PEP8 (specifically, no underscores in package/module names). That is done in two commits: 86d45cba557680f6198e95ab1507b55dd8f2678a for packaging and 8714903c93e05bcbc534e3638cad58ac294ccaaa for underscores in file names.

PR #76, opened 2016-01-07
Refactor packaging

fnothaft:
This is just the two commits that refactor the package structure of toil-scripts from #73. @hannes-ucsc suggested that we split these out into a separate PR that can merge quickly, as to not block development, which I agree with. I am still working to address the following two comments that @hannes-ucsc made on #73:

> 3. User scripts (or rather user modules) must be invoked with python -m followed by a fully-qualified module path, e.g. ...

> 5. Some of the module and package names are a bit long, maybe we can remove the _pipeline from the end of module or package names. We could also adopt the convention of always naming user modules toil.py or main.py. The above example would then become ...


hannes-ucsc:
@jvivian, are you ok with this? It will break your wrapper scripts!

jvivian:
@hannes-ucsc The fix will be changing `python <script>.py` to `python -m <script>.py` in the wrappers, correct?  This will cause a bunch of merge conflicts for me, but I don't mind dealing it.  Better to just treat it like a band-aid and merge now.

hannes-ucsc:
@jvivian Not quite, the current directory needs to be src, the -m arg needs to be a fully qualified module path, not a relative script file path. See https://github.com/BD2KGenomics/toil-scripts/pull/73#issuecomment-169592209 for details.

hannes-ucsc:
@jvivian Instead of changing cwd to src, you can also make sure that the absolute path to src is in PYTHONPATH. I think I'll just go ahead and fix the wrapper scripts. In a continuation PR.

fnothaft:
@hannes-ucsc if you want to open a PR against my fork/branch with the module path changes, then we can get this all in a single PR.

hannes-ucsc:
Continued in #77.

PR #83, opened 2016-01-14
Lock down the ADAM version.

fnothaft:
Depends on https://github.com/BD2KGenomics/cgl-docker-lib/pull/70.

### DIFF
--- a/src/toil_scripts/adam_pipeline/spark_toil_script.py
+++ b/src/toil_scripts/adam_pipeline/spark_toil_script.py
@@ -1,3 +1,4 @@
+
 #!/usr/bin/env python2.7
 """
 UCSC Computational Genomics Lab RNA-seq Pipeline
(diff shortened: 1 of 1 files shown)


heuermh:
+1

fnothaft:
Actually, I think I'm going to need to bump the version of ADAM. Let's hold on this 1 sec. Ran into something when testing last night.

fnothaft:
Specifically, I'd like to get https://github.com/bigdatagenomics/adam/pull/921 in. @heuermh can you review/merge on ADAM and I'll bump here and at cgl-docker-lib?

heuermh:
Merged #921, go ahead

hannes-ucsc:
Should we create a special, toil-specific branch in the ADAM repo and refer to that branch instead of a specific commit sha1 in the Dockerfile? That branch would get merged into from master whenever you want the container to be updated. Then you just rebuild the containers. That way we don't need to update this Dockerfile everytime a relevant update is made to ADAM. I'm sure their will be more fixes.

fnothaft:
Updated.

heuermh:
I'm not sure whether we should keep bumping the commit, use a toil-specific branch, or cut a 0.19.0 release of ADAM.  What is the reason not to use head at this point?

fnothaft:
I'd like to cut a 0.19.0... [once we've got the requisite issues closed](https://github.com/bigdatagenomics/adam/milestones/0.19.0).

I prefer a commit to latest/a toil branch in ADAM because if you fetch the commit and it fails, then you get an error message, viz https://github.com/BD2KGenomics/cgl-docker-lib/pull/71. I see this as a feature, not a bug --> it ensures that we're all running the exact same code.

heuermh:
Agreed.

@jpdna

PR #86, opened 2016-01-18
Added alt-aware alignment for GRCh38 reference build to batch alignment pipeline

fnothaft:
Adds the [bwa-postalt](https://github.com/BD2KGenomics/cgl-docker-lib/pull/73) step to the batch alignment pipeline. This step [lifts alt alignments](https://github.com/lh3/bwa/blob/master/README-alt.md) back to the non-alt contigs and is necessary for finalizing alt aware read alignments.

Also, changes the bwa pipeline to lock down to bwa 0.7.12.

I've tagged this `needs work` as I still need to test this end-to-end.


src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -308,19 +339,54 @@ def run_bwa(job, job_vars):
...
+    with open(os.path.join(work_dir, output), 'w') as samfile:
+        docker_call(tool='quay.io/ucsc_cgl/bwa:0.7.12',
                     tool_parameters=parameters, work_dir=work_dir, outfile=samfile, sudo=sudo)
+    return job.fileStore.writeGlobalFile(os.path.join(work_dir, output))
+
+
+def lift_alts(job, job_vars, sam_id):
+    """
+    This lifts alt haplotype alignments back to the reference contigs that the
+    alts vary from. This is necessary for alt aware alignment against GRCh38.
```
heuermh: Could you clarify, is this necessary because bwa is being called without the ```-a``` parameter?  Are alignments copied over?  Many alignments might be dropped since whole genes (e.g. HLA-DRB5 on chr 6) and even larger regions (A vs B haplotypes in KIR region on chr 19) are not present in the reference chromosomes.

jvivian:
Sadly, we have no continuous integration setup for toil-scripts, so @fnothaft just ping me once you're ready to merge (if you don't have privileges).

fnothaft:
Closing in favor of #95.

PR #82, opened 2016-01-14
Mount volume to Spark containers to use as a local directory for shuffles

fnothaft:
Relies on https://github.com/BD2KGenomics/cgl-docker-lib/pull/69.

@almussel can you test this on your current cluster?

PR #94, opened 2016-01-21
Mount volume to Spark containers to use as a local directory for shuf…

hannes-ucsc:
…fles.

Relies on https://github.com/BD2KGenomics/cgl-docker-lib/pull/69.

Rebase of #82


fnothaft:
Thanks @hannes-ucsc !

PR #96, opened 2016-01-25
Adam pipeline bug fixes

almussel:


src/toil_scripts/adam_pipeline/spark_toil_script.py
```diff
@@ -162,9 +187,11 @@ def adam_transform(job, masterIP, inFile, snpFile, inputs):
                 "--conf", "spark.hadoop.fs.default.name=hdfs://%s:%s" % (masterIP, HDFS_MASTER_PORT),
                 "--conf", "spark.local.dir=/ephemeral/spark",
                 "--", "transform", 
-                "mkdups.adam", "ri.adam",
+                "hdfs://"+masterIP+":"+HDFS_MASTER_PORT+"/mkdups.adam", "hdfs://"+masterIP+":"+HDFS_MASTER_PORT+"/ri.adam",
```
fnothaft: Can we split this line into two for readability?

Also, perhaps:

```
"hdfs://%s:%s/mkdups.adam" % (masterIP, HDFS_MASTER_PORT)
```

Instead of:

```
"hdfs://"+masterIP+":"+HDFS_MASTER_PORT+"/mkdups.adam"
```

src/toil_scripts/adam_pipeline/spark_toil_script.py
```diff
@@ -173,9 +200,11 @@ def adam_transform(job, masterIP, inFile, snpFile, inputs):
                 "--conf", "spark.hadoop.fs.default.name=hdfs://%s:%s" % (masterIP, HDFS_MASTER_PORT),
                 "--conf", "spark.local.dir=/ephemeral/spark",
                 "--", "transform", 
-                "ri.adam", "bqsr.adam",
+                "hdfs://"+masterIP+":"+HDFS_MASTER_PORT+"/ri.adam", "hdfs://"+masterIP+":"+HDFS_MASTER_PORT+"/bqsr.adam",
```
fnothaft: Ditto here --> split line, use string formatter.

src/toil_scripts/adam_pipeline/spark_toil_script.py
```diff
@@ -241,6 +272,7 @@ def stop(self):
         log.write("stop masters\n")
         log.flush()
 
+        call(["docker", "exec", self.sparkContainerID, "rm", "-r", "/ephemeral/spark"])
```
fnothaft: I think it'd be preferable to stop the spark container, and to then shutil.rmtree("/mnt/ephemeral/spark"). I'm just imagining that something uncouth could happen if we yanked the directory out while Spark was still running.

src/toil_scripts/adam_pipeline/spark_toil_script.py
```diff
@@ -285,8 +317,10 @@ def stop(self):
         log.write("stop workers\n")
         log.flush()
 
+        call(["docker", "exec", self.sparkContainerID, "rm", "-r", "/ephemeral/spark"])
```
fnothaft: Ditto here.

src/toil_scripts/adam_pipeline/spark_toil_script.py
```diff
@@ -285,8 +317,10 @@ def stop(self):
...
         call(["docker", "stop", self.sparkContainerID])
         call(["docker", "rm", self.sparkContainerID])
+        call(["docker", "exec", self.hdfsContainerID, "rm", "-r", "/ephemeral/hdfs"])
```
fnothaft: And here too.

fnothaft:
Dropped comments inline. LGTM after comments are addressed. Thanks for the fixes, @almussel !

almussel:
I cleaned up the formatting like you suggested, but I don't think removing the directories after shutdown will work. Docker runs as root, which means the directories it creates are only removable by an administrator or from within the container. What sort of bad things do you imagine happening if the directory is deleted after all the jobs are done running?

fnothaft:
> Docker runs as root, which means the directories it creates are only removable by an administrator or from within the container. What sort of bad things do you imagine happening if the directory is deleted after all the jobs are done running?

Ah... thanks for pointing that out. I always forget the permissions angle with the `/mnt` volumes in cgcloud. I don't think anything terrible would happen (although HDFS may notice data loss and try to recover from that), it's just cleaner to remove the directory after shutdown. Both the HDFS and Spark images contain shutdown scripts. Perhaps let's keep this as you've done for now, and we can do a refactor later that runs the shutdown scripts and removes the work directories.

fnothaft:
Looks good for merge from my side. Thanks @almussel!

hannes-ucsc:
Most docker containers use this [1] wrapper to deal with the root-owned files and directories. It fixes the permissions without hard-coding the uid of the desired owner. For it to work, the parent directory needs to be owned by the desired owner which implies that the parent directory is created by the toil-script.

[1]: https://github.com/BD2KGenomics/cgl-docker-lib/blob/master/rsem-1.1.13/wrapper.sh

almussel:
@hannes-ucsc do you think we should emulate that here? To me it seems cleaner to delete the directories from within docker, since we already have permission there.

hannes-ucsc:
Talked to @almussel offline. She's going to try a slightly different approach to setting the work and local directories for Spark.

fnothaft:
@hannes-ucsc @jvivian We are good to go here. Can we get a merge?

heuermh:
My tests with a simpler pipeline was based on this, +1 from me

src/toil_scripts/adam_pipeline/spark_toil_script.py
```diff
@@ -74,9 +74,21 @@ def call_conductor(masterIP, inputs, src, dst):
...
+                "--conf", "spark.executor.memory=%s" % inputs["executorMemory"],
                 "--", "-C", src, dst])
 
 
+def remove_file(masterIP, filename):
+    """
+    Remove the given file from hdfs with master at the given IP address
+    """
+    containerID = check_output(["ssh", "-o", "StrictHostKeyChecking=no", masterIP, "docker", "ps", \
+                                "|", "grep", "apache-hadoop-master", "|", "awk", "'{print $1}'"])[:-1]
```
hannes-ucsc: For small outputs it may be easier and more readable to do the filtering and column extraction in Python rather than piping through grep and awk.

PR #114, opened 2016-02-24
ADAM überscript (resolves #102)

### LINKED ISSUE #102: Create Überscript for ADAM/Toil test
- Wrap cgcloud create
- Metric collection process
- Launch all-samples toil script (either ADAM or GATK)

almussel:
Resolves #102


almussel:
Resolves issue #102

fnothaft:
Looking...

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,408 @@
...
+metric_start_time_margin = 1800
+
+scaling_initial_wait_period_in_seconds = 300
+cluster_scaling_interval_in_seconds = 300
+
+cluster_size_lock = threading.Lock()
+
+aws_region = 'us-west-2'
+
+ref = "https://s3-us-west-2.amazonaws.com/cgl-pipeline-inputs/variant_grch38/GRCh38_full_analysis_set_plus_decoy_hla.fa"
```
fnothaft: We're going to have different reference input sets for GRCh38 vs HG19. It might be good to have this in a dictionary instead of in a bunch of globals. This would have the additional benefit that we could swap out the dictionaries easily, IMO.

fnothaft: Also, these links are out of date. Can you change to the `variant_grch38_reordered` bucket?

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,408 @@
...
+                           '--cluster-name', params.cluster_name,
+                           '--spot-bid', str(params.spot_bid),
+                           '--leader-on-demand',
+                           '--num-threads', str(params.num_workers),
+                           '--ssh-opts',
+                           '-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no',
+                           'toil'])
+
+
+def place_boto_on_leader(params):
```
fnothaft: Don't need to fix this for the recompute, but after we're done, let's allow the user to supply a `.aws/credentials` file as well. I've opened #115 to track this.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,408 @@
...
+                               "--ref", ref,
+                               "--amb" amb,
+                               "--ann", ann,
+                               "--bwt", bwt,
+                               "--pac", pac,
+                               "--sa", sa,
+                               "--fai", fai,
+                               "--alt", alt,
+                               "--use_bwakit",
+                               "--num_nodes", "2", # what does this number affect? find out
```
fnothaft: This is the size of the launched Spark cluster, _including_ the master node.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,408 @@
...
+                               "--driver_memory", "50g", #entire node
```
fnothaft: This should be an argument.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,408 @@
...
+                               '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no',
+                               'screen', '-dmS', params.cluster_name])
+        # Run command on screen session
+        subprocess.check_call(['cgcloud', 'ssh', '--cluster-name', params.cluster_name, 'toil-leader',
+                               '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no',
+                               'screen', '-S', params.cluster_name, '-X', 'stuff',
+                               "python", "-m", "toil-scripts.adam_gatk_pipeline.align_and_call",
+                               "aws:{0}:{1}".format(aws_region, jobstore),
+                               "--retryCount", "1",
+                               "--s3_bucket", "fnothaft-fc-test-west-2", # MAKE THIS A VARIABLE
```
fnothaft: This should be an argument.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,408 @@
...
+                               "--executor_memory", "50g",#entire node
+                               "--phase", phase,
+                               "--mills", mills,
+                               "--dbsnp", dbsnp,
+                               "--omni", omni,
+                               "--hapmap", hapmap,
+                               "--batchSystem", "mesos",
+                               "--mesosMaster", "$(hostname -i):5050",
+                               "--workDir", "/var/lib/toil"
+                               "--file_size", "1G"
```
fnothaft: This should be an argument.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,408 @@
...
+        # Create screen session
+        subprocess.check_call(['cgcloud', 'ssh', '--cluster-name', params.cluster_name, 'toil-leader',
+                               '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no',
+                               'screen', '-dmS', params.cluster_name])
+        # Run command on screen session
+        subprocess.check_call(['cgcloud', 'ssh', '--cluster-name', params.cluster_name, 'toil-leader',
+                               '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no',
+                               'screen', '-S', params.cluster_name, '-X', 'stuff',
+                               "python", "-m", "toil-scripts.adam_gatk_pipeline.align_and_call",
+                               "aws:{0}:{1}".format(aws_region, jobstore),
```
fnothaft: Where's the sample manifest file injected from?

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,408 @@
...
+                               "--mills", mills,
+                               "--dbsnp", dbsnp,
+                               "--omni", omni,
+                               "--hapmap", hapmap,
+                               "--batchSystem", "mesos",
+                               "--mesosMaster", "$(hostname -i):5050",
+                               "--workDir", "/var/lib/toil"
+                               "--file_size", "1G"
+                               "--logInfo",
+                               restart])
```
fnothaft: Maybe s/`,\n\srestart])`/`] + [restart])`\g?

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,408 @@
...
+    subparsers = parser.add_subparsers(dest='command')
+
+    # Launch Cluster
+    parser_cluster = subparsers.add_parser('launch-cluster', help='Launches AWS cluster via CGCloud')
+    parser_cluster.add_argument('-s', '--num-workers', required=True, help='Number of workers desired in the cluster.')
+    parser_cluster.add_argument('-c', '--cluster-name', required=True, help='Name of cluster.')
+    parser_cluster.add_argument('-S', '--share', required=True,
+                                help='Full path to directory: pipeline script, launch script, config, and master key.')
+    parser_cluster.add_argument('--spot-bid', default=1.00, help='Change spot price of instances')
+    parser_cluster.add_argument('-t', '--instance-type', default='c3.8xlarge',
```
fnothaft: Let's default to m3.8xlarge.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,408 @@
...
+def main():
+    """
+    Modular script for running toil pipelines
+    """
+    parser = argparse.ArgumentParser(description=main.__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
+    subparsers = parser.add_subparsers(dest='command')
+
+    # Launch Cluster
+    parser_cluster = subparsers.add_parser('launch-cluster', help='Launches AWS cluster via CGCloud')
+    parser_cluster.add_argument('-s', '--num-workers', required=True, help='Number of workers desired in the cluster.')
```
fnothaft: Is this the number desired in the cluster at launch?

fnothaft:
I just made my first review pass. Can we squash these 3 commits down to 1?

almussel:
For sure. Once I've incorporated your suggestions and cleaned things up I'll squash everything down.

src/toil_scripts/adam_uberscript/adam_uberscript.py
hannes-ucsc: s/bucket/directory/

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,408 @@
...
+        subprocess.check_call(['cgcloud', 'ssh', '--cluster-name', params.cluster_name, 'toil-leader',
+                               '-o', 'UserKnownHostsFile=/dev/null', '-o', 'StrictHostKeyChecking=no',
+                               'screen', '-S', params.cluster_name, '-X', 'stuff',
+                               "python", "-m", "toil-scripts.adam_gatk_pipeline.align_and_call",
+                               "aws:{0}:{1}".format(aws_region, jobstore),
+                               "--retryCount", "1",
+                               "--s3_bucket", "fnothaft-fc-test-west-2", # MAKE THIS A VARIABLE
+                               "--bucket_region", aws_region,
+                               "--aws_access_key", "${FC_AWS_ACCESS_KEY_ID}", # THESE SHOULDN'T BE HERE
+                               "--aws_secret_key", "${FC_AWS_SECRET_ACCESS_KEY}", # THESE SHOULDN'T BE HERE
```
hannes-ucsc: Please remove.

src/toil_scripts/adam_uberscript/adam_uberscript.py
hannes-ucsc: +1

src/toil_scripts/adam_uberscript/automated_scaling.py
```diff
@@ -0,0 +1,212 @@
...
+            version = int(nodes_per_sample.pop('version'))
+        else:
+            #dom.put_attributes("nodes_per_sample", {'version': str(0)})
+            version = 0
+        samples = OrderedDict(map(lambda x:[x[0], map(int, x[1].split(','))],
+                                  sorted(nodes_per_sample.items(),
+                                         key=(lambda x: x[1])))) 
+        return cls(conn, dom, version, samples)
+    
+    def update(self):
```
hannes-ucsc: Eliminate

src/toil_scripts/adam_uberscript/automated_scaling.py
```diff
@@ -0,0 +1,212 @@
...
+                                    expected_value=('version', str(self.version)))
+            self.reorder()
+            return True
+        except boto.exception.SDBResponseError as e:
+            if e.error_code == 'ConditionalCheckFailed':
+                return False
+            else:
+                raise e
+    
+    def increase_nodes(self, sampleID, n):
```
hannes-ucsc: Make static

src/toil_scripts/adam_uberscript/automated_scaling.py
```diff
@@ -0,0 +1,212 @@
...
+                sum += nodes
+                if sample == sampleID and sum <= cluster_size.size:    
+                    return
+                if sum > cluster_size:
+                    break
+            self.update()
+            cluster_size.update()
+            time.sleep(60)
+
+    def decrease_nodes(self, sampleID, n):
```
hannes-ucsc: Make static

src/toil_scripts/adam_uberscript/automated_scaling.py
```diff
@@ -0,0 +1,212 @@
...
+            dom.put_attributes("cluster_size", {'version': str(0)})
+            version = 0
+        if 'size' in cluster_size:
+            size = int(cluster_size['size'])
+        else:
+            dom.put_attributes("cluster_size", {'size': str(0)})
+            size = 0
+        return cls(conn, dom, version, size)
+
+    def update(self):
```
hannes-ucsc: Eliminate

src/toil_scripts/adam_uberscript/automated_scaling.py
```diff
@@ -0,0 +1,212 @@
...
+        try:
+            self.dom.put_attributes("cluster_size", attributes, expected_value=('version', str(self.version)))
+            return True
+        except boto.exception.SDBResponseError as e:
+            if e.error_code == 'ConditionalCheckFailed':
+                return False
+            else:
+                raise
+
+    def change_size(self, n):
```
hannes-ucsc: Make static

hannes-ucsc: or @classmethod

fnothaft:
Here are the current operating instructions for getting set up and running with the überscript:

```
1. Create the head node:
    - cgcloud create -i m3.medium -n / generic-ubuntu-trusty-box
2. On the head node:
    - sudo apt-get install git emacs make libxml2-dev libxslt1-dev python-dev
    - virtualenv cgcloud-master
    - source cgcloud-master/bin/activate
    - pip install boto tqdm
    - Add to bash profile:
        - export CGCLOUD_ZONE=us-west-2a
        - export CGCLOUD_PLUGINS=cgcloud.mesos:cgcloud.toil
        - export CGCLOUD_KEYPAIRS="__me__ @@developers"
        - export CGCLOUD_NAMESPACE=/ 
    - git clone https://github.com/BD2KGenomics/cgcloud.git
    - from the cgcloud repository:
        - make develop sdist
    - create a .boto
    - git clone https://github.com/BD2KGenomics/toil-scripts.git
    - checkout adam_uberscript branch
3. Running the pipelines:
    These should be run in succession, in a screen, from the head node. 
    a. Launch Cluster:
        PYTHONPATH=$PYTHONPATH:~/toil-scripts/src python -m toil_scripts.adam_uberscript.adam_uberscript launch-cluster -c <cluster name> -S <toil scripts directory> -t r3.8xlarge -b ~/.boto
    b. Launch Pipeline:
        PYTHONPATH=$PYTHONPATH:~/toil-scripts/src python -m toil_scripts.adam_uberscript.adam_uberscript launch-pipeline -c <cluster name> -j <jobstore> -B <data bucket> -m <memory per worker node> -s <spark cluster size including master>
    c. Launch Metrics:
        PYTHONPATH=$PYTHONPATH:~/toil-scripts/src python -m toil_scripts.adam_uberscript.adam_uberscript launch-metrics -c "small-test" -j <jobstore> --namespace $CGCLOUD_NAMESPACE
```

fnothaft:
I'm getting an error that seems to indicate that we don't have a toil-box built yet in the FireCloud `/` namespace, so I'm building that now.

fnothaft:
Copying from @jpdna's email to @hannes-ucsc:

> I'm not clear how cgcloud namespaces work and what their scope is.
> I see here:
> "cgcloud create -t m3.medium -n / generic-ubuntu-trusty-box"
>
> the namespaces being created as as "/"
>
>Should I be using "/" in my tests of Audrey's uberscript, or could that conflict with another user - like Frank?
>
> Should  I use "-n justin_uber_test"  
>
>I also see in Audrey's instructions: " export CGCLOUD_NAMESPACE=/ "
> so surely will need to make sure it matches ,and maybe elsewhere...

hannes-ucsc:
* namespaces do for images, instances and clusters what directories do for files on a file system
* their names always begin and end with a slash (unlike directory)
* they can be nested as in `/foo/bar/` but the nesting is largely inconsequential, IOW. `/foo/bar/` is just as isolated from `/foo/` as `/bar/` would be.
* the default namespace is `/__me__/` where
* `__me__` is a placeholder for A) your IAM user name minus the @ and anything thereafter or B) the value of the `CGCLOUD_ME` environment variable
* any IAM user can switch to any namespace at any time. Namespaces have no impact on permissions or access. They are merely an organizational tool
* the `CGCLOUD_NAMESPACE` variable sets the default namespace
* the `-n` command line option overrides the default for a single invocation of cgcloud

By using the / namespace you indicate that you are creating a cluster shared by the team. But again, this is just a convention. It would not be a problem if you did everything in `/jpdna/`, just slightly more confusing. Confusing for us, but mostly for you since you might also want to run an independent, personal experiment/cluster in that namespace.

hannes-ucsc:
@fnothaft 

> I'm getting an error that seems to indicate that we don't have a toil-box built yet in the FireCloud / namespace, so I'm building that now.

Keep in mind that the resulting image will likely differ between versions of cgcloud. An image created with   cgcloud 1.3.x from the stable branch will likely not work (or worse, work subtly different) when used with cgcloud 1.4.x from the master branch. For this run, we should be using a development checkout of the master branch. This is what Audrey included in her instructions.

fnothaft:
> Keep in mind that the resulting image will likely differ between versions of cgcloud. An image created with cgcloud 1.3.x from the stable branch will likely not work (or worse, work subtly different) when used with cgcloud 1.4.x from the master branch. For this run, we should be using a development checkout of the master branch. This is what Audrey included in her instructions.

Ah, good point. Thanks for pointing this out. I did it from a checkout of the master branch on my home machine, but this is about a week and a half out of date.

fnothaft:
Did someone just accidentally terminate the wrong machine? My headnode got shut down...

hannes-ucsc:
Wasn't me.

jpdna:
It was me, accidentally, by running:
cgcloud terminate -n / generic-ubuntu-trusty-box
which was also the name of mine.
So clearly we can't both use the "/" namespace.

I will not use "/".

fnothaft:
> So clearly we can't both use the "/" namespace.
>
> I will not use "/".

Sounds like a plan; I will abandon "/" as well...

hannes-ucsc:
You can both use the / namespace. But it's a shared namespace so you need to coordinate. If you're just running tests, go ahead and use the personal namespace. But the real 10-node run should be in the root namespace.

fnothaft:
> But the real 10-node run should be in the root namespace.

What exactly is the specific benefit to running in the root namespace vs. any other namespace? I think for now we're just doing testing, so it's a bit of a moot point. We'd definitely want to start a clean cluster for the final run. I'm more interested in the philosophical reasoning behind running in the root namespace.

hannes-ucsc:
The image will be created in that namespace. If we need to rerun something there will be no second guessing as to whose namespace was used. To me this is a no-brainer: you each do transient experiments in your personal namespaces but the real thing is run at the root or at any other namespace whose name indicates the importance and shared nature of that namespace. 

Since this is so obvious to me I'm also puzzled as to why you were both using the root namespace. Were you both redundantly trying to do the real run? 

Anyways, whoever is driving the bus uses the / namespace and creates the head node in it. They also run the commands on it that spawn the cluster and kick off the pipeline. Other people can then ssh into the head and leader nodes, watch the screens, look at log files etc.

fnothaft:
> Since this is so obvious to me I'm also puzzled as to why you were both using the root namespace. Were you both redundantly trying to do the real run?

No, I think we are both unfamiliar with cgcloud's namespace feature and were blindly following the scripts. In any case, I get the sentiment now. We probably wouldn't want to run the final experiment in `/`, but in something like `/adam-recompute`, since `/` would presumably be shared with the RNA-seq experiments, no?

hannes-ucsc:
Yes, much better.

fnothaft:
As a small warning, you can't make your namespace name too long. E.g., `/frank-uberscript-test/` causes an AWS instance naming error (longer than 64 chars) when creating the toil-box image. I should've copied down the error message, but alas, I just terminated the instance, so it is gone to the wind.

fnothaft:
TODO: Disable strict host checking during ssh. I had a run just fail due to a ssh timeout at the host checking prompt when I left the room for a few minutes.

fnothaft:
Actually, it isn't an ssh, it's an [rsync](https://github.com/BD2KGenomics/toil-scripts/pull/114/files#diff-7060afbea50ff85e89fe57836bc27314R70) call that already appears to disable host checking. I'll need to look at that more later...

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -262,6 +262,9 @@ def download_shared_files(job, input_args):
     """
     shared_files = ['ref.fa', 'ref.fa.amb', 'ref.fa.ann', 'ref.fa.bwt', 'ref.fa.pac', 'ref.fa.sa', 'ref.fa.fai']
 
+    if "autoscale_cluster" in input_args and input_args["autoscale_cluster"]:
```
fnothaft: I think this code is accidentally included.

fnothaft:
The pipeline script is terminating immediately because the manifest file and toil_scripts repos are getting rsynced to slightly different paths vs. what we expected. Also, there's [some code that accidentally got tracked into bwa_alignment](https://github.com/BD2KGenomics/toil-scripts/pull/114/files#diff-fb1369a3679ce478d903128ad802b9e9R265) that seems to be giving us the business. I've just commented it out for now.

fnothaft:
Also, what is a bit unusual is that we rsync the toil_scripts dir over on cluster launch, not before each run. This means that if you update toil scripts on the headnode, it doesn't update on the cluster. This is just a bit confusing from the user perspective.

fnothaft:
Otherwise though, it looks like we're off to the races! I will report back with more news sometime later.

fnothaft:
The job just failed because of a configuration issue on my side, so I am restarting.

As an aside, @hannes-ucsc, it looks like `cgcloud ssh` from the headnode to the toil leader has broken. I can manually ssh between the two, but `cgcloud ssh` is giving me no love, but worse yet, no error message beyond "ssh failed":

```
~/toil-scripts/src$ cgcloud ssh --cluster-name fan-us-toil-test toil-leader                                                                                  
INFO: Using zone 'us-west-2a' and namespace '/fan-us-test/'
INFO: Binding to instance ... 
INFO: ... waiting for instance i-7dadcdba ... 
INFO: ... running, waiting for assignment of public IP ... 
INFO: ... assigned, waiting for SSH port ... 
INFO: ... open ... 
INFO: ... instance ready.
Permission denied (publickey).
  File "/home/ubuntu/cgcloud-master/bin/cgcloud", line 9, in <module>
    load_entry_point('cgcloud-core', 'console_scripts', 'cgcloud')()
  File "/home/ubuntu/cgcloud/core/src/cgcloud/core/cli.py", line 49, in main
    app.run( args )
  File "/home/ubuntu/cgcloud/lib/src/cgcloud/lib/util.py", line 306, in run
    command.run( options )
  File "/home/ubuntu/cgcloud/core/src/cgcloud/core/commands.py", line 81, in run
    return self.run_in_ctx( options, ctx )
  File "/home/ubuntu/cgcloud/core/src/cgcloud/core/commands.py", line 105, in run_in_ctx
    return self.run_on_role( options, ctx, role )
  File "/home/ubuntu/cgcloud/core/src/cgcloud/core/commands.py", line 124, in run_on_role
    return self.run_on_box( options, box )
  File "/home/ubuntu/cgcloud/core/src/cgcloud/core/commands.py", line 187, in run_on_box
    self.run_on_instance( options, box )
  File "/home/ubuntu/cgcloud/core/src/cgcloud/core/commands.py", line 252, in run_on_instance
    status = self.ssh( options, box )
  File "/home/ubuntu/cgcloud/core/src/cgcloud/core/commands.py", line 241, in ssh
    return box.ssh( user=self._user( box, options ), command=options.command )
  File "/home/ubuntu/cgcloud/core/src/cgcloud/core/box.py", line 1284, in ssh
    raise RuntimeError( 'ssh failed' )
```

fnothaft:
Hrm, it appears that I did something that caused two toil-leaders to come into existance, which is probably why the cgcloud ssh failed. I'm recreating the toil cluster and restarting the run.

fnothaft:
And, something had gone wrong with the ssh service on my headnode. Re-adding my ssh keys seems to have solved the issue.

fnothaft:
Messed up and put a bucket in the `us-east-1`... Moving my data back to `us-west-2`. That issue will be fixed when #118 merges with @hannes-ucsc's changes from #119/https://github.com/fnothaft/toil-scripts/pull/6.

hannes-ucsc:
One thing to keep in mind is that SSH agent forwarding doesn't work well with screen.

https://gist.github.com/martijnvermaat/8070533

So running `ssh` or `cgcloud ssh` in a reattached screen on the head node will fail if you are relying on agent forwarding for authentication. To fix, we should use a dedicated key pair on the head node. I've modified the above instructions (look for *new*).

In any case, don't put your own private keys on the head node.

hannes-ucsc:
The image creation step is missing from the instructions. Adding that, too.

fnothaft:
> In any case, don't put your own private keys on the head node.

I've done that for now, since I'm just testing, but yes, I agree for the production run.

I'll read up on the link you sent later this PM.

fnothaft:
Dropping a note to myself here: we probably want to change where the scale up command is issued from. We're issuing it at the point that we schedule the Spark master service job, but I think we want to issue it before we fork the ADAM/GATK pipelines, because I _think_ you can deadlock a single node cluster otherwise.

fnothaft:
Noone touched my headnode, right? I logged in this morning and my tmux sessions were dead...?

fnothaft:
For now, I am working off of a copy of the [adam_überscript branch in my fork](https://github.com/fnothaft/toil-scripts/tree/adam_uberscript) with the following modifications:

* Rebased on #118 
* https://github.com/fnothaft/toil-scripts/commit/4b8cdb7607aaf70883a25b5b077092b6d3add133 consolidates all of the scale up/down calls into the `adam_gatk_pipeline`. I think this is cleaner than having the autoscaling code spread across all of the contingent pipelines, since we only need to change cluster size at the boundary of the various pipelines.
* https://github.com/fnothaft/toil-scripts/commit/5064b096fa8949e3bc97e29f059aa0b291bcc030 makes a variety of fixes in the überscript to resolve miscellaneous issues described [in a comment above](https://github.com/BD2KGenomics/toil-scripts/pull/114#issuecomment-189586488)

fnothaft:
As a note, you'll need to run `cgcloud create -IT toil-box` every time you register a new keypair. I've updated that above.

fnothaft:
Just pushed a few more fixes to my fork:

* https://github.com/fnothaft/toil-scripts/commit/5f5c3f8c9091ed2e7ff5258a3c29bfc0d6645622 makes the manifest file a command line input, not a hardcoded parameter
* https://github.com/fnothaft/toil-scripts/commit/14fc17a8b94bdcca3788de592965cfa382b68366, https://github.com/fnothaft/toil-scripts/commit/f79273ed8d1f7b83434e367797f403a980d674cd, and https://github.com/fnothaft/toil-scripts/commit/65863fb592206a54e5daf5bdd50c10a40aba1621 fix miscellaneous syntax/etc issues

hannes-ucsc:
> Noone touched my headnode, right?

Not me.

> fnothaft@4b8cdb7 consolidates all of the scale up/down calls into the adam_gatk_pipeline. I think this is cleaner than having the autoscaling code spread across all of the contingent pipelines, since we only need to change cluster size at the boundary of the various pipelines.

Won't that create 10 idle Spark nodes while alignment is running? @almussel pushed the scaling down to avoid wasting money on idle nodes.

fnothaft:
Pushed some more fixes at my fork (https://github.com/BD2KGenomics/toil-scripts/commit/0d9bd0378fff26ea8a05fe9a3f5f6a3c38ca8a6d, https://github.com/BD2KGenomics/toil-scripts/commit/1dd530f190d708d80ff132eae8fe6046983eac0c, https://github.com/BD2KGenomics/toil-scripts/commit/edc79d17bb0fd4e744c9de5b3eb65da7f184b9b4, https://github.com/BD2KGenomics/toil-scripts/commit/9bab0ddcbf2c3d20ae2fda5b87a01f64726dc27e), mostly just syntax stuff. This has made it through the first round of scaling up, so w00t! I'll update as the script runs on.

fnothaft:
> Won't that create 10 idle Spark nodes while alignment is running? @almussel pushed the scaling down to avoid wasting money on idle nodes.

Nyet! To highlight the relevant sections of the code, we create scale up ([allocate preprocessing nodes](https://github.com/fnothaft/toil-scripts/commit/4b8cdb7607aaf70883a25b5b077092b6d3add133#diff-924a4449d5efb0928b3441ac334f0c85R378)) and scale down ([post-ADAM](https://github.com/fnothaft/toil-scripts/commit/4b8cdb7607aaf70883a25b5b077092b6d3add133#diff-924a4449d5efb0928b3441ac334f0c85R390)) jobs. We then conditionally (e.g., for post-alignment, see [here](https://github.com/fnothaft/toil-scripts/commit/4b8cdb7607aaf70883a25b5b077092b6d3add133#diff-924a4449d5efb0928b3441ac334f0c85R387) and [here](https://github.com/fnothaft/toil-scripts/commit/4b8cdb7607aaf70883a25b5b077092b6d3add133#diff-924a4449d5efb0928b3441ac334f0c85R414) and [here](https://github.com/fnothaft/toil-scripts/commit/4b8cdb7607aaf70883a25b5b077092b6d3add133#diff-924a4449d5efb0928b3441ac334f0c85R438)) weave them into our DAG. This way, we don't have to touch any of the pipelines, nor do we need to keep machines turned on unnecessarily.

Also, I added code to do a scale-up during the [sample loop](https://github.com/fnothaft/toil-scripts/commit/4b8cdb7607aaf70883a25b5b077092b6d3add133#diff-924a4449d5efb0928b3441ac334f0c85R247). This was missing from the original code, but I think is necessary to make autoscaling work. Specifically, if you start with one node and wait until BWA is running to increment the cluster size, BWA schedules on a full machine (by requesting all the cores), so you'll only ever get one sample through at a time. If you bump the cluster size _before_ scheduling any of the per-sample compute, you won't block like this.

Also, I just realized a possible issue with the core count requests... @hannes-ucsc you'd mentioned this a while back, but we're requesting the full core count of a machine by using the `multiprocessing` library to query the core count on the current machine... Since we're running these clusters with a small toil-leader, I'm not sure how that'll play out. Although, I think that all of the invocations of the `multiprocessing` library occur inside of toil `Job`s and thus should be invoked on a `toil-worker`. TBD; but I think we're fine. I'm just musing out loud.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,409 @@
...
+        cluster_size = get_cluster_size()
+        desired_cluster_size = get_desired_cluster_size()
+        if cluster_size < desired_cluster_size:
+            with node_allocation_lock:
+                grow_cluster(desired_cluster_size - cluster_size, params.instance_type, cluster_name, spot_bid)
+        update_cluster_size(desired_cluster_size)
+
+        # Sleep
+        resize_time = time.time() - size_check_time
+        log.info('Cluster is {} nodes as of {}'.format(get_cluster_size(), resize_time))
```
fnothaft: Small logging irregularity here. I've opened #121 to track this; it is not important to fix now, but we should fix it later.

fnothaft:
Just pushed a couple more commits (https://github.com/fnothaft/toil-scripts/commit/085087c231478094bd522bef530615b51099c569 and https://github.com/fnothaft/toil-scripts/commit/9a4418d285a3f1c61daf22f2344128d64f7aab95) that fix typos I injected when consolidating the autoscaling code into the `adam_gatk_pipeline`, and I rebased on top of @hannes-ucsc's merge of #118.

fnothaft:
Pushed https://github.com/BD2KGenomics/toil-scripts/commit/aae5d5023209d94d8704df3d01910775a20da7d1 which fixes an issue where the instance type wasn't known to the metrics monitor. I've also opened #122 to discuss whether it would be good to automatically detect the cluster worker instance type.

fnothaft:
I still didn't want to force push to replace this branch, so I've opened #123. This seems to run end-to-end OK on my test sample. I've spun up a cluster to run NA12878 through ADAM and GATK for validation. The namespace is `/adam-na12878/` for those interested in following along. Can I get a final review @almussel and @hannes-ucsc on #123? I would like to get it reviewed and merged by EOD tomorrow, and then we can start the 10 sample benchmarking recompute.

fnothaft:
Not sure what happened there, but I lost my tmux session on the head node for `/adam-na12878/`... I've shuttered the cluster and will look more in the AM.

hannes-ucsc:
> lost my tmux session

This was the thing I wanted to mention during scrum: until you figure out what's killing your tmux sessions, lets stick with screen. We've never lost a screen session on our head nodes before.

hannes-ucsc:
We seem to have moved on to #123. Closing this one.

PR #119, opened 2016-02-26
Add support for downloading s3:// URLs using Boto across all pipelines

hannes-ucsc:
plus cosmetics

PR #120, opened 2016-02-27
Add support for downloading s3:// URLs using Boto across all pipelines

hannes-ucsc:
plus cosmetics

Rebased version of #119.


hannes-ucsc:
Sorry, wrong target.

PR #123, opened 2016-02-29
Überscript, squashed

fnothaft:
This is a continuation of #114, with bugfixes, squashed down, docs moved inline, etc.

### DIFF
--- a/src/toil_scripts/adam_uberscript/adam_uberscript.py
+++ b/src/toil_scripts/adam_uberscript/adam_uberscript.py
@@ -72,7 +72,7 @@ def launch_cluster(params):
                            '--ssh-opts="-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"',
                            'toil-leader',
                            '-a',
-                           inputs['manifest'], ":~/manifest"])
+                           params.manifest_path, ":~/manifest"])
     subprocess.check_call(['cgcloud',
                            'rsync',
                            '--zone', "{0}a".format(aws_region),
@@ -234,6 +234,7 @@ def grow_cluster(n, instance_type, cluster_name, spot_bid):
                                '--cluster-name', str(cluster_name),
                                '--spot-bid', str(spot_bid),
                                '--zone', ("%sa" % aws_region),
+                               '--spot-tentative',
                                'toil'])
 
         # how much did we grow by?
@@ -417,6 +418,7 @@ def main():
     parser_cluster.add_argument('-T', '--leader-type', default='m3.medium', help='Sets leader instance type.')
     parser_cluster.add_argument('-b', '--boto-path', default='/home/mesosbox/.boto', type=str,
                                 help='Path to local .boto file to be placed on leader.')
+    parser_cluster.add_argument('-M', '--manifest-path', required=True, help='Path to manifest file.')
 
     # Launch Pipeline
     parser_pipeline = subparsers.add_parser('launch-pipeline', help='Launches pipeline')
--- a/src/toil_scripts/adam_uberscript/adam_uberscript.py
+++ b/src/toil_scripts/adam_uberscript/adam_uberscript.py
@@ -5,17 +5,23 @@ Date: 2/12/16
 
 Modified from John Vivian's automated_scaling_tests.py script
 """
+
 import logging
+
+# Initialize logging before the remaining imports to prevent those imported modules from snatching that one shot at 
+# basicConfig.
+
 log = logging.getLogger(__name__)
 logging.basicConfig(level=logging.INFO,
                     format='%(asctime)-15s:%(levelname)s:%(name)s:%(message)s',
                     datefmt='%m-%d %H:%M:%S')
 
+from StringIO import StringIO
+import operator
 import argparse
 from collections import namedtuple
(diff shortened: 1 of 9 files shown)


fnothaft:
Action items from today's call:

* Remove dependency on boto/etc in `adam_gatk_pipeline`
* Clean up spurious diffs in final PR in tertiary pipelines
* Remove job encapsulation
* Remove argument validation for jobstore

@hannes-ucsc will add something about spot tentative

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,457 @@
...
+        log.info('Growing cluster by {} nodes of type: {}'.format(diff, instance_type))
+
+        # call to cgcloud to increase the cluster size
+        subprocess.check_call(['cgcloud',
+                               'grow-cluster',
+                               '--list',
+                               '--instance-type', str(instance_type),
+                               '--num-workers', str(diff),
+                               '--cluster-name', str(cluster_name),
+                               '--spot-bid', str(spot_bid),
```
hannes-ucsc: Add --spot-tentative here.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -0,0 +1,457 @@
...
+                               '--zone', ("%sa" % aws_region),
```
hannes-ucsc: Big piece of functionality missing here. We need to balance between the availability zones.

I'll file a PR against this one.

hannes-ucsc:
Everyone, read second bullet of https://github.com/BD2KGenomics/toil/wiki/Python-Dos-and-Don'ts.

hannes-ucsc:
And don't do `from foo import *` unless foo has an __all__

All hell breaks loose if you do anyways because you will end up importing what foo imports.

fnothaft:
> Everyone, read second bullet of https://github.com/BD2KGenomics/toil/wiki/Python-Dos-and-Don'ts.

Ah, good to know. I haven't used python logging libraries much before, so I didn't realize they were variadic.

> And don't do `from foo import *` unless `foo` has an `__all__`

Which part of the code is this in reference to? The `adam_gatk_pipeline`? That code path doesn't seem to work with explicit imports due to what I think is some sort of serialization issue I've mentioned earlier.

fnothaft:
>> And don't do `from foo import *` unless `foo` has an `__all__`
>
> Which part of the code is this in reference to? The adam_gatk_pipeline? That code path doesn't seem to work with explicit imports due to what I think is some sort of serialization issue I've mentioned earlier.

Nevermind; figured it out from https://github.com/fnothaft/toil-scripts/pull/7.

fnothaft:
Pushed an update with the fixes discussed earlier today. I will test and remove the `needs work` tag.

fnothaft:
Just pushed a [new commit](https://github.com/fnothaft/toil-scripts/commit/90c510ddc679e3d0d6cc98c77c9da7825ef27e51) with additional fixes from testing tonight. Mostly around error checking on the `cgcloud grow-cluster` call. Also, there's something that I think is an off-by-one error RE: the desired cluster size. I've got these fixes running now and will report back later as to whether they are sufficient.

fnothaft:
Also, I'm not thrilled with how spot instance grant times are blowing up with the `--spot-tentative` flag and the 5 minute backoff and retry. This seems to make the process of waiting for spot instances get really long. How well did this approach work on the RNA-seq recompute?

fnothaft:
Hrmmm...

So, something we didn't consider: what do we do when running a spot instance based cluster, and we lose the _Spark master_? We just had that happen on my test cluster, and alas, as is designed now, I am afraid the cluster is unrecoverable because there is no way to change the master IP that the workers are looking at.

Argh!!!

I'm going to take my test cluster down for now. We'll need to think about this further...

fnothaft:
Also, the `--spot-tentative` flag can't be working right. I just logged into the AWS console and... we have 95 machines running, most of which are spot instances that we gave up on.

fnothaft:
@hannes-ucsc I've just pushed the commit we discussed.

hannes-ucsc:
> Also, I'm not thrilled with how spot instance grant times are blowing up with the --spot-tentative flag and the 5 minute backoff and retry. This seems to make the process of waiting for spot instances get really long. How well did this approach work on the RNA-seq recompute?

`--quick` and `--spot-tentative` should make `cgcloud grow-cluster` return much quicker. We can discuss the 5 minute wait. The reasoning is that unless all instances were provisioned, there is no need to retry to provision the rest of them immediately.

hannes-ucsc:
For TCGA, we were adding a 100 instances at a time. If the grow_cluster() function is called for each sample then that might slow overall provisioning down. But I don't think it is.

fnothaft:
> For TCGA, we were adding a 100 instances at a time. If the grow_cluster() function is called for each sample then that might slow overall provisioning down. But I don't think it is.

It is being called per sample. I think it should be straightforward to batch.

fnothaft:
Ping @hannes-ucsc can we get this merged or did you want any more changes?

Issue #132, opened 2016-03-03
Modify überscript to run HG19 and GRCh38


fnothaft:
@jpdna has reheadered the HG19 files, which means that we can now run the adam-vs-gatk überscript pipeline on both HG19 and GRCh38.

@jpdna can you:

* Add a file similar to https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/adam_uberscript/input_files.py for HG19
* Modify the [überscript](https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/adam_uberscript/adam_uberscript.py) to select between GRCh38 and HG19?

Once you open a PR with these changes, I, @hannes-ucsc, and @almussel will review. Once it looks good, we can kick off the 10 sample HG19 run. I'll add AIs for this on #113.

jpdna:
PR #133 resolves this hopefully.

@fnothaft  - I want to now do a test run of this on the cluster.    Doing so should not  create any conflict with your runs given we are using seperate CGCLOUD namespaces, but wanted to double check with you before I start as if there is any danger  I'd wait.

fnothaft:
Reopening this; @jpdna @hannes-ucsc the permissions for `s3://cgl-pipeline-inputs/variant_hg19_reordered` are screwy. We were using `https://` URLs, but the files aren't public and the AWS console won't let me make them public. I've [added a commit](https://github.com/fnothaft/toil-scripts/commit/591f63e2da0c0fac63b661fb4635a9b8b8dae74f) to #171 to WAR this, but it'd be good to have a second set of eyes on the issue.

hannes-ucsc:
Probably ACLs that get in the way.

fnothaft:
> Probably ACLs that get in the way.

Actually, the odd thing is that it looks like it is more the _absence_ of ACLs that is getting in the way.

fnothaft:
This was resolved by #159.

PR #133, opened 2016-03-03
Select between GRCh38 and HG19

jpdna:
Allows selection between `GRCh38` and `hg19` using the `--reference-genome` parameter
as discussed in #132

This code has only been tested for parameter parsing, it has not yet been run on a cluster, let me know if you want me to test it further before this PR is accepted

### DIFF
--- a/src/toil_scripts/adam_uberscript/adam_uberscript.py
+++ b/src/toil_scripts/adam_uberscript/adam_uberscript.py
@@ -125,28 +125,29 @@ def launch_pipeline(params):
                                '-o', 'StrictHostKeyChecking=no',
                                'screen', '-dmS', params.cluster_name])
 
+        # Run command on screen session  
 
-        # do we have a defined master ip?
-        masterIP_arg = ""
-        if params.master_ip:
-            masterIP_arg = "--master_ip %s" % params.master_ip
+        if params.reference_genome == 'GRCh38':
+            from toil_scripts.adam_uberscript.input_files import GRCh38_inputs as inputs
+        elif params.reference_genome == 'hg19':
+            from toil_scripts.adam_uberscript.input_files import hg19_inputs as inputs
+        else:
+            raise Exception("Invalid reference genome, reference should have been  validated as a choice during argument parsing.  Should not have reached this point, arg parsing code likely out of date.")
 
-        # Run command on screen session        
-        pipeline_command = ("PYTHONPATH=$PYTHONPATH:~/toil-scripts/src python -m toil_scripts.adam_gatk_pipeline.align_and_call " +
-                            "aws:{region}:{j} " +
-                            "--autoscale_cluster " +
-                            "--sequence_dir {sequence_dir} " +
-                            "--retryCount 1 " +
-                            "--s3_bucket {b} " +
-                            "--bucket_region {region} " +
-                            "--uuid_manifest ~/manifest " +
-                            "--ref {ref} " +
-                            "--amb {amb} " +
-                            "--ann {ann} " +
-                            "--bwt {bwt} " +
-                            "--pac {pac} " +
-                            "--sa {sa} " +
-                            "--fai {fai} " ) 
+        pipeline_command = "PYTHONPATH=$PYTHONPATH:~/toil-scripts/src python -m toil_scripts.adam_gatk_pipeline.align_and_call " + \
+                           "aws:{region}:{j} " + \
+                           "--autoscale_cluster " + \
+                           "--retryCount 1 " + \
+                           "--s3_bucket {b} " + \
+                           "--bucket_region {region} " + \
+                           "--uuid_manifest ~/manifest " + \
+                           "--ref {ref} " + \
+                           "--amb {amb} " + \
+                           "--ann {ann} " + \
+                           "--bwt {bwt} " + \
+                           "--pac {pac} " + \
+                           "--sa {sa} " + \
+                           "--fai {fai} "
         if 'alt' in inputs:
           pipeline_command +=   pipeline_command += "--alt {alt} "
         pipeline_command += "--use_bwakit " +
--- a/src/toil_scripts/adam_uberscript/adam_uberscript.py
+++ b/src/toil_scripts/adam_uberscript/adam_uberscript.py
@@ -37,7 +37,6 @@ import boto.sdb
 from boto.ec2 import connect_to_region
 
 from automated_scaling import ClusterSize, Samples
-from toil_scripts.adam_uberscript.input_files import inputs
 
 metric_endtime_margin = timedelta(hours=1)
 metric_initial_wait_period_in_seconds = 0
@@ -126,6 +125,7 @@ def launch_pipeline(params):
                                '-o', 'StrictHostKeyChecking=no',
                                'screen', '-dmS', params.cluster_name])
 
+
         # do we have a defined master ip?
         masterIP_arg = ""
         if params.master_ip:
@@ -146,9 +146,10 @@ def launch_pipeline(params):
                             "--bwt {bwt} " +
                             "--pac {pac} " +
                             "--sa {sa} " +
-                            "--fai {fai} " +
-                            "--alt {alt} " +
-                            "--use_bwakit " +
+                            "--fai {fai} " ) 
+        if 'alt' in inputs:
+          pipeline_command +=   pipeline_command += "--alt {alt} "
+        pipeline_command += "--use_bwakit " +
                             "--num_nodes {s} " +
                             "--driver_memory {m} " +
                             "--executor_memory {m} " +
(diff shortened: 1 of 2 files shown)


fnothaft:
LGTM!

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -119,8 +118,13 @@ def launch_pipeline(params):
                                '-o', 'UserKnownHostsFile=/dev/null',
                                '-o', 'StrictHostKeyChecking=no',
                                'screen', '-dmS', params.cluster_name])
-        # Run command on screen session
-        
+        # Run command on screen session  
+
+        if params.reference_genome == 'GRCh38':
+            from toil_scripts.adam_uberscript.input_files import GRCh38_inputs as inputs
+        if params.reference_genome == 'hg19':
```
hannes-ucsc: elif

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -119,8 +118,13 @@ def launch_pipeline(params):
...
+            from toil_scripts.adam_uberscript.input_files import hg19_inputs as inputs
```
hannes-ucsc: add else branch raising exception?

src/toil_scripts/adam_uberscript/input_files.py
```diff
@@ -1,5 +1,5 @@
 
-inputs = {    "ref": "https://s3-us-west-2.amazonaws.com/cgl-pipeline-inputs/variant_grch38_reordered/GRCh38_full_analysis_set_plus_decoy_hla.reordered.fa",
+GRCh38_inputs = {    "ref": "https://s3-us-west-2.amazonaws.com/cgl-pipeline-inputs/variant_grch38_reordered/GRCh38_full_analysis_set_plus_decoy_hla.reordered.fa",
               "amb": "https://s3-us-west-2.amazonaws.com/cgl-pipeline-inputs/variant_grch38_reordered/GRCh38_full_analysis_set_plus_decoy_hla.reordered.fa.amb",
```
hannes-ucsc: indentation seems off

src/toil_scripts/adam_uberscript/input_files.py
```diff
@@ -12,3 +12,19 @@
             "dbsnp": "https://s3-us-west-2.amazonaws.com/cgl-pipeline-inputs/variant_grch38_reordered/ALL_20141222.dbSNP142_human_GRCh38.snps.vcf",
              "omni": "https://s3-us-west-2.amazonaws.com/cgl-pipeline-inputs/variant_grch38_reordered/1000G_omni2.5.grch38.reordered.vcf",
            "hapmap": "https://s3-us-west-2.amazonaws.com/cgl-pipeline-inputs/variant_grch38_reordered/hapmap_3.3.grch38.reordered.vcf"}
+
+hg19_inputs = {    "ref": "https://s3-us-west-2.amazonaws.com/cgl-pipeline-inputs/variant_hg19_reordered/hg19.fa",
```
hannes-ucsc: ditto

src/toil_scripts/adam_uberscript/adam_uberscript.py
fnothaft: This gets validated during argument parsing. See https://github.com/BD2KGenomics/toil-scripts/pull/133/files#diff-7060afbea50ff85e89fe57836bc27314R451.

jpdna: > This gets validated during argument parsing. See https://github.com/BD2KGenomics/toil-scripts/pull/133/files#diff-7060afbea50ff85e89fe57836bc27314R451.
@hannes-ucsc - let me know if you still advise changes here.

hannes-ucsc: I would still use elif and have an else: branch that asserts False. That's just basic defensive programming. Of course, *right now* Frank is right, but *over time* his assertion may break.

fnothaft: My strong preference is to validate arguments as early as is possible; my understanding is that by setting the `choices` param when argparsing, we push the validation to when the arguments are parsed and cannot get invalid inputs here. I'd prefer to document this decision in the argparse section of the code, and to not revalidate here.

Also, this is personal preference, but I prefer to `assert` if an internal invariant fails, and to raise an exception (e.g., `IllegalArgumentException` in Java, `ValueError` in Python) if a user provides bad input. I find `assert False` to be a smell; if you're `assert False`ing, you should almost always `raise Exception("informative error message")` instead. In my world, if an `assert` fires, that indicates that there is a bug in the program that should be fixed, while raising an exception to the user tells them that they need to fix their inputs.

hannes-ucsc: Yes, validate early and raise RuntimeException to alert user but **also** assert here. If someone changes the validation code and forgets to change this code here, they will have introduced a bug which would be caught by the assert. This is especially important considering Python's lack of type-safe enumerations. And -1 on informational message, the stack trace will be informative enough.

fnothaft: > Yes, validate early and raise RuntimeException to alert user but also assert here. If someone changes the validation code and forgets to change this code here, they will have introduced a bug which would be caught by the assert.

Just wanted to point out that the flip side of this is that if someone adds a new reference build, then they need to modify validation code in two places. My preference is to validate the inputs in a single place, and to validate them as early as is possible.

> And -1 on informational message, the stack trace will be informative enough.

Are you actually _-1_ on this? It doesn't make sense to me to require the end user to parse out a stack trace when you can raise an informative error message instead.

hannes-ucsc: After talking to @fnothaft, this is the compromise: add `elif` and `else` with `assert False, "unknown reference genome %s" % params.reference_genome`.

hannes-ucsc:
I want clean the überscript up a bit so I will go ahead and rebase and merge this. Otherwise, I'd be causing even more conflicts.

@jpdna I use SourceTree and XCode's FileMerge for conflict resolution. PyCHarm also has an excellent GUI for merge conflict resolution.

jpdna:
I just anyhow  pushed the re-based version to my repo, and so will update
the pull request anyhow, so you can start from wherever you prefer Hannes -
thanks!

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -126,29 +125,32 @@ def launch_pipeline(params):
                                '-o', 'StrictHostKeyChecking=no',
                                'screen', '-dmS', params.cluster_name])
 
-        # do we have a defined master ip?
-        masterIP_arg = ""
-        if params.master_ip:
-            masterIP_arg = "--master_ip %s" % params.master_ip
```
hannes-ucsc: Did you intent to remove this?

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -126,29 +125,32 @@ def launch_pipeline(params):
...
+                           "--uuid_manifest ~/manifest " + \
+                           "--ref {ref} " + \
+                           "--amb {amb} " + \
+                           "--ann {ann} " + \
+                           "--bwt {bwt} " + \
+                           "--pac {pac} " + \
+                           "--sa {sa} " + \
+                           "--fai {fai} "
+        if 'alt' in inputs:
+          pipeline_command +=   pipeline_command += "--alt {alt} "
```
hannes-ucsc: we use four spaces for indent

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -126,29 +125,32 @@ def launch_pipeline(params):
...
-                            "--alt {alt} " +
-                            "--use_bwakit " +
+        # Run command on screen session  
+
+        if params.reference_genome == 'GRCh38':
+            from toil_scripts.adam_uberscript.input_files import GRCh38_inputs as inputs
+        elif params.reference_genome == 'hg19':
+            from toil_scripts.adam_uberscript.input_files import hg19_inputs as inputs
+        else:
+            raise Exception("Invalid reference genome, reference should have been  validated as a choice during argument parsing.  Should not have reached this point, arg parsing code likely out of date.")
```
hannes-ucsc: This is way too verbose and doesn't contain the value that could not be matched.

hannes-ucsc:
Thanks, I'll take it from here.

src/toil_scripts/adam_uberscript/adam_uberscript.py
hannes-ucsc: And it's got and extra `pipeline_command += ` in there

jpdna: Indeed, I'll check more carefully, thanks Hannes

hannes-ucsc:
Superceded by #159

Issue #141, opened 2016-03-04
Make s3am block size parametrizable


fnothaft:
Where given, s3am block size is hardcoded. E.g., https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/rnaseq_cgl/rnaseq_cgl_pipeline.py#L873. This is a low priority.

fnothaft:
Copying @hannes-ucsc's comment from #141:

> This will do for now but it would be nice to have three retries as in
>
> https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/rnaseq_cgl/rnaseq_cgl_pipeline.py#L873

Even better: we should pull every s3am call out into a library call that wraps the retry count, encryption, etc.

hannes-ucsc:
Use 64MB

jpdna:
> Even better: we should pull every s3am call out into a library call that wraps the retry count, encryption, etc.
@hannes-ucsc - should I put this s3am wrapper library function in toil_scripts/__init__.py similar to https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/__init__.py#L5  ?

hannes-ucsc:
That library call exists now. It should default to passing 64MB  for `--part-size`.

jvivian:
Won't fix — hardcoded as 50MB in common library and @hannes-ucsc deems that sufficient.

PR #142, opened 2016-03-04
Set s3am block size and parallelism in batch_alignment.

### LINKED ISSUE #140: s3am block size is too small in alignment pipeline
Specifically, it isn't set, and thus s3am defaults to a 5MB block size. Since a file is limited to 10k blocks, this limits file size to ~50GB.

We should do something like https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/rnaseq_cgl/rnaseq_cgl_pipeline.py#L873 and up the block size to 64M. We should also set the upload parallelism.

fnothaft:
Resolves #140. Sets s3am block size to 64M; this should be made parametrizable at a later date. I've opened #141 to track. Also, sets the number of cores to use on upload, and fixes a missing UUID.

### DIFF
--- a/src/toil_scripts/batch_alignment/bwa_alignment.py
+++ b/src/toil_scripts/batch_alignment/bwa_alignment.py
@@ -536,6 +536,9 @@ def upload_to_s3(work_dir, input_args, output_file):
     # what is the path we are uploading to?
     s3_path = os.path.join('s3://', bucket_name, bucket_dir, output_file)
 
+    # how many cores should we use?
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
This will do for now but it would be nice to have three retries as in 

https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/rnaseq_cgl/rnaseq_cgl_pipeline.py#L873

Issue #139, opened 2016-03-03
Sort Simons FASTQ files


jpdna:
The BAM->FASTQ conversion using `samtools fastq -1 /data/output_1.fastq -2 /data/output_2.fastq -0 /data/output_0.fastq /data/mybam.bam` resulted in fastq files which were not sorted, so pairs did not line up between the two files.

I'm testing now sorting them using:
```
zcat LP6005441-DNA_A01_1.fastq.gz | paste - - - - | sort -k1,1 -S 10G -T temp/ |  tr '\t' '\n' | gzip > /sorted/LP6005441-DNA_A01_1.fastq.gz 
```
and if successful will run this sorting as a toil job, and update original BAM->FASTQ conversion code to sort by default.
I'll provide an example pair of files from a single sample ASAP for use in testing.

fnothaft:
Thanks @jpdna! Indeed, BWA-MEM is processing the current files as single ended reads:

```
[M::mem_process_seqs] Processed 3200000 reads in 1058.497 CPU sec, 34.540 real sec
[M::process] read 3200000 sequences (320000000 bp)...
[M::process] 3200000 single-end sequences; 0 paired-end sequences
[M::mem_process_seqs] Processed 3200000 reads in 1072.141 CPU sec, 34.810 real sec 
```

fnothaft:
Ah, also, can you make sure to get the NA12878 sample? We'll need that for the first recompute.

jpdna:
> Ah, also, can you make sure to get the NA12878 sample? We'll need that for the first recompute

yup

jpdna:
you will find now the first sorted pair of fastq files for a single sample at:
`s3://cgl-sgdd-work/sequence-sorted`
The read pairs  in the sorted file do  all match one-to-one on row basis as expected.  The files are confirmed to have the correct (same) number of rows post sorting.
I ran the first 2500 paired-reads (head -n 10000) through bwa mem like 
```
bwa mem -t 12 GRCh38_full_analysis_set_plus_decoy_hla.reordered.fa LP6005441-DNA_A01_1.sorted.fastq.head10000 LP6005441-DNA_A01_2.sorted.fastq.head10000
```
and output does refer to pairs, so all seems good.
I'm starting a big run shortly and the sorted files will begin appearing at `s3://cgl-sgdd-work/sequence-sorted`
@fnothaft - I will run NA12878 first and let you know when it is done and where to find

jpdna:
[...]

PR #148, opened 2016-03-07
Allow users to be able to specify directory where raw sequences live.

fnothaft:
We will want this since #139 moves the location of the FASTQ files in our bucket.

### DIFF
--- a/src/toil_scripts/adam_gatk_pipeline/align_and_call.py
+++ b/src/toil_scripts/adam_gatk_pipeline/align_and_call.py
@@ -148,6 +148,11 @@ def build_parser():
                         help = "Skip preprocessing and start running from variant calling. Implies --skip_alignment.",
                         action = 'store_true')
 
+    # what directory are our sequence files in?
+    parser.add_argument('-SD', '--sequence_dir',
+                        help = 'Directory where raw sequences are.',
+                        default = 'sequence')
+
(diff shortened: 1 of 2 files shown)


src/toil_scripts/adam_gatk_pipeline/align_and_call.py
```diff
@@ -148,6 +148,11 @@ def build_parser():
                         help = "Skip preprocessing and start running from variant calling. Implies --skip_alignment.",
                         action = 'store_true')
 
+    # what directory are our sequence files in?
+    parser.add_argument('-SD', '--sequence_dir',
```
hannes-ucsc: Unix convention is that `-SD` is the contraction of `-S -D`. I would either use --sd or ditch it completely.

fnothaft: Yeah, I think we're at the point where the short arguments aren't very useful. However, we've got a lot of them in here. Would it be OK with you to hold off on changing this for now, and to then go back after we're done and refactor them all? If so, I'll open a ticket to track.

hannes-ucsc: Absolutely.

hannes-ucsc:
+1

But see https://github.com/BD2KGenomics/toil-scripts/issues/150

Issue #151, opened 2016-03-08
picard AddOrReplaceReadGroups fails unpredictibly


fnothaft:
For whatever reason, when using [Picard](https://github.com/BD2KGenomics/cgl-docker-lib/tree/master/picardtools) to attach ReadGroups to our aligned files, this occasionally fails and instead of writing out the same BAM file as we started with, but with the new read group attached, it writes out a 30KB corrupt BAM file. This is obviously disappointing.

- [x] @fnothaft will WAR this using `bwa-kit`, which is actually what we should do anyways because rewriting a 150GB BAM file to add ~30B of ASCII text to the header is wasteful. (see #153)
- [ ] @almussel will debug _why_ this is happening with Picard? You can repro it by running the batch alignment pipeline on `LP6005441-DNA_A12` from the SGD project. (this issue)
- [ ] We should fix the garbage we're writing in the platform fields.
- [ ] We should close the gaps between the bwa and bwakit pipelines (see #154)

fnothaft:
@jvivian is there a reason we're writing garbage metadata into the headers? I'm pretty sure not all data is being sequenced on [machine 12345](https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/batch_alignment/bwa_alignment.py#L511).

jvivian:
@fnothaft Yes. GATK _requires_ the PU field. The samples I was initially aligning did not have sequence machine info and that was a placeholder that never got put into argparse (my fault).  Should I fix that on a different branch?

```
These parameters need to be used

RGLB=String
LB=String                     Read Group Library  Required.

RGPU=String

PU=String                     Read Group platform unit (eg. run barcode)  Required.

RGSM=String
SM=String                     Read Group sample name  Required.
```

fnothaft:
> @fnothaft will WAR this using bwa-kit, which is actually what we should do anyways because rewriting a 150GB BAM file to add ~30B of ASCII text to the header is wasteful.

#153

fnothaft:
> @fnothaft Yes. GATK requires the PU field.

SRSLY? AH MAH ZING.

> The samples I was initially aligning did not have sequence machine info and that was a placeholder that never got put into argparse (my fault). Should I fix that on a different branch?

Ah yes, this is thorny now. Let me put a junk PU into #153.

[...]

PR #153, opened 2016-03-08
Use bwakit to write out read group info in bwakit pipeline.

fnothaft:
In the bwakit pipeline, we are seeing odd Picard failures (see #151). However, in the bwakit pipeline, we can write the @RG lines when we write our initial BAM file. This commit makes that change, and eliminates the further DAG stages.

CC @jvivian for review
I'll open a ticket for rationalizing the differences between the BWA and BWAKIT pipelines.

Tagged this with `needs work` while I am doing the test run.

### DIFF
--- a/src/toil_scripts/batch_alignment/bwa_alignment.py
+++ b/src/toil_scripts/batch_alignment/bwa_alignment.py
@@ -335,17 +335,13 @@ def static_dag_declaration(job, job_vars):
     input_args, ids = job_vars
     cores = input_args['cpu_count']
 
-    # if we run with bwakit, we skip conversion
+    # if we run with bwakit, we skip conversion and reheadering
     if input_args['use_bwakit']:
 
         bwa = job.wrapJobFn(run_bwa, job_vars, cores=cores, disk=input_args['file_size'])
-        header = job.wrapJobFn(fix_bam_header, job_vars, bwa.rv(), disk=input_args['file_size'])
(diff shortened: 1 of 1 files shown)


src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -382,10 +395,19 @@ def run_bwa(job, job_vars):
         return_input_paths(job, work_dir, ids, 'ref.fa.alt')
 
     if input_args['use_bwakit']:
+
+        # add a read group line
+        # this is an illumina only pipeline, therefore, hardcoding illumina is OK
+        uuid = input_args['uuid']
+        library = input_args['lb']
+        rg = "@RG\tID:%s\tLB:%s\tPL:ILLUMINA\tSM:%s" % (uuid,
```
jvivian: Do you not need a PU tag here?

fnothaft: Saw your comment on #151 and updated this. Thanks for pointing that out. It's a definite hack but... GD GATK, amirite?

fnothaft:
This is good to go; `needs work` tag is removed.

hannes-ucsc:
+1 but would like to give @jvivian a chance to weigh in

fnothaft:
> +1 but would like to give @jvivian a chance to weigh in
SGTM!

fnothaft:
Ping @jvivian for review/merge.

PR #157, opened 2016-03-08
Auto-scaling of standalone Spark cluster (resolves #152)

### LINKED ISSUE #152: Auto-scale external Spark cluster
https://github.com/BD2KGenomics/toil-scripts/issues/130#issuecomment-191557158

hannes-ucsc:
Resolves #152 

We now distinguish between *dynamic* ("Spark-on-Toil") and *static* clusters. Within the latter we further distinguish between *standalone* and *external*. A standalone cluster is managed by the überscript. This PR obsoletes the `--add-to-etc-hosts` parameter and replaces it with `--master_ip=auto` to indicate standalone clusters. That special string is resolved in the Toil scripts using SimpleDB (the SparkMasterAddress model) and injected into the driver containers via Docker's `--add-host` option. For that I had to dig a little deeper into my OO magic bag of trick. Apologies for that.

This PR creates the standalone Spark master when it is needed. This has the advantage that it will be recreated if needed and that it does not sit idle during alignment.

This has not been tested. The doctests pass, though, needless to say.

### DIFF
--- a/src/toil_scripts/adam_pipeline/adam_preprocessing.py
+++ b/src/toil_scripts/adam_pipeline/adam_preprocessing.py
@@ -24,37 +24,76 @@ Toil pipeline for ADAM preprocessing
 ================================================================================
 :Dependencies
 docker          - apt-get install docker (or 'docker.io' for linux)
-toil            - pip install --pre toil
+toil            - pip install toil
 """
 
-# python core libraries
 import argparse
 import logging
 import multiprocessing
 import os
-from subprocess import call, check_call, check_output
+from subprocess import check_call, check_output
 import sys
-import time
 
-# toil imports
 from toil.job import Job
 
-# toil scripts imports
+from toil_scripts.adam_uberscript.automated_scaling import SparkMasterAddress
 from toil_scripts.batch_alignment.bwa_alignment import docker_call
-from toil_scripts.spark_utils.spawn_cluster import *
+from toil_scripts.spark_utils.spawn_cluster import start_spark_hdfs_cluster
 
 SPARK_MASTER_PORT = "7077"
 HDFS_MASTER_PORT = "8020"
 log = logging.getLogger(__name__)
 
+
+class MasterAddress(str):
+    """
+    A string containing the hostname or IP of the Spark/HDFS master. The Spark master expects its own address to
+    match what the client uses to connect to it. For example, if the master is configured with a host name,
+    the driver can't use an IP address to connect to it, and vice versa. This class works around by distinguishing
+    between the notional master address (self) and the actual one (self.actual) and adds support for the special
+    master address "auto" in order to implement auto-discovery of the master of a standalone.
+
+    >>> foo = MasterAddress('foo')
+    >>> foo == 'foo'
+    True
+    >>> foo.actual == 'foo'
+    True
+    >>> foo.actual == foo
+    True
+    """
+    def __init__(self, master_ip):
+        # TBD: this could do more tricks like always mapping an IP address to 'spark-master' etc.
+        if master_ip == 'auto':
+            super(MasterAddress, self).__init__('spark-master')
+            self.actual = SparkMasterAddress.load_in_toil().value
+        else:
+            super(MasterAddress, self).__init__(master_ip)
+            self.actual = self
+
+    def docker_parameters(self, docker_parameters=None):
+        """
+        Augment a list of "docker run" arguments with those needed to map the notional Spark master address to the
+        real one, if they are different.
+        """
+        if self != self.actual:
+            add_host_option = '--add-host=spark-master:' + self.actual
+            if docker_parameters is None:
+                docker_parameters = [add_host_option]
+            else:
+                docker_parameters.append(add_host_option)
+        return docker_parameters
+
+
 def call_conductor(masterIP, inputs, src, dst):
     """
-    Invokes the conductor container.
+    Invokes the Conductor container to copy files between S3 and HDFS
+
+    :type masterIP: MasterAddress
     """
     docker_call(no_rm = True,
                 work_dir = os.getcwd(),
                 tool = "quay.io/ucsc_cgl/conductor",
-                docker_parameters = ["--net=host"],
+                docker_parameters = masterIP.docker_parameters(["--net=host"]),
                 tool_parameters = ["--master", "spark://"+masterIP+":"+SPARK_MASTER_PORT,
                  "--conf", "spark.driver.memory=%sg" % inputs["driverMemory"],
                  "--conf", "spark.executor.memory=%sg" % inputs["executorMemory"],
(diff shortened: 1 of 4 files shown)


src/toil_scripts/adam_uberscript/automated_scaling.py
```diff
@@ -149,6 +150,161 @@ def change_size(cls, conn, dom, n):
...
+                self.value -= delta
+                self.version += 1
+                if self._update():
+                    break
+                else:
+                    yield attempt, True
+
+    def release(self, delta=1):
+        """
+        Repeatedly attempt to decrease the semaphore by the given amount until successful.
```
hannes-ucsc: "increase"

hannes-ucsc:
@fnothaft, could you take a peek?

src/toil_scripts/adam_pipeline/adam_preprocessing.py
```diff
@@ -24,37 +24,75 @@
...
+            super(MasterAddress, self).__init__(master_ip)
+            self.actual = self
+
+    def docker_parameters(self, docker_parameters=None):
+        """
+        Augment a list of "docker run" arguments with those needed to map the notional Spark master address to the
+        real one, if they are different.
+        """
+        if self != self.actual:
+            add_host_option = '--add-host=spark-master:' + self.actual
```
fnothaft: Ah, savvy! Good call.

src/toil_scripts/adam_pipeline/adam_preprocessing.py
```diff
@@ -81,36 +122,40 @@ def call_adam(masterIP, inputs, arguments):
 def remove_file(masterIP, filename, sparkOnToil):
     """
     Remove the given file from hdfs with master at the given IP address
+
+    :type masterIP: MasterAddress
     """
+    masterIP = masterIP.actual
     if sparkOnToil:
-        containerID = check_output(["ssh", "-o", "StrictHostKeyChecking=no", masterIP, "docker", "ps", \
+        containerID = check_output(["ssh", "-o", "StrictHostKeyChecking=no", masterIP, "docker", "ps",
```
fnothaft: Small nit: if I'm splitting a call/array across multiple lines, I'll typically break it out into one line per element/logical group of elements. E.g., this would become something like:

```
containerID = check_output(["ssh",
"-o", "StrictHostKeyChecking=no",
masterIP,
"docker", "ps", "|",
"grep", "apache-hadoop-master", "|",
"awk", "'{print $1}'"])
```

Same for the line below.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -220,110 +218,179 @@ def get_cluster_size(cluster_name):
...
+            num_samples = params.spark_sample_slots - num_unused_sample_slots
+            assert num_samples >= 0
+            if num_samples:
+                # ... At least one sample is. How many Spark workers are needed for all of them.
+                spark_workers_per_sample = params.spark_nodes - 1
+                num_desired_workers = num_samples * spark_workers_per_sample
+                num_workers = len(list_nodes(params.cluster_name, role='spark-slave'))
+                num_masters = len(list_nodes(params.cluster_name, role='spark-master'))
+                if num_masters == 0:
+                    # Absence of master indicates absence of cluster, so create it now.
```
fnothaft: Ah, nice! So, this will _auto start_ the Spark cluster as well?

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -220,110 +218,179 @@ def get_cluster_size(cluster_name):
...
+                    assert num_workers == 0, 'Orphaned Spark workers detected'
+                    output = check_output(['cgcloud',
+                                           'create-cluster',
+                                           '--list',
+                                           '--zone', aws_region + 'a',
```
fnothaft: Do we want to hard code this?

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -220,110 +218,179 @@ def get_cluster_size(cluster_name):
...
+                                           '--cluster-name', params.cluster_name,
+                                           '--leader-instance-type', params.spark_master_type or params.leader_type,
+                                           '--instance-type', params.spark_instance_type or params.instance_type,
+                                           '--num-workers', str(num_desired_workers),  # TODO: spot options
```
fnothaft: Make a ticket to track for later?

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -220,110 +218,179 @@ def get_cluster_size(cluster_name):
...
+                    SparkMasterAddress.create(domain, master['private_ip_address']).save(force=True)
+                elif num_masters == 1:
+                    # There already is a Spark cluster, so scale it if necessary. 
+                    if num_desired_workers > num_workers:
+                        grow_cluster(num_nodes=num_desired_workers - num_workers,
+                                     instance_type=params.instance_type,
+                                     cluster_name=params.cluster_name,
+                                     cluster_type='spark')
+                    elif num_desired_workers < num_workers:
+                        pass  # TODO: think about scaling down
```
fnothaft: Can't we reuse the toil cluster shrinking code?

fnothaft:
Generally LGTM, other than a few nits. Do we want to merge this before we have autoshrink implemented?

src/toil_scripts/adam_pipeline/adam_preprocessing.py
hannes-ucsc: Agreed. The grep and awk pipe should be replaced with pure Python, too.

src/toil_scripts/adam_uberscript/adam_uberscript.py
hannes-ucsc: Yes.

src/toil_scripts/adam_uberscript/adam_uberscript.py
hannes-ucsc: No. I'll add a parameter for this.

src/toil_scripts/adam_uberscript/adam_uberscript.py
hannes-ucsc: To remove instances, sure. But what about the HDFS contents? Don't we need to to some rebalancing before taking instances out of service?

src/toil_scripts/adam_uberscript/adam_uberscript.py
hannes-ucsc: https://github.com/BD2KGenomics/toil-scripts/issues/174

hannes-ucsc:
@fnothaft, I'd prefer to merge ASAP (after I make the changes you proposed, except auto-shrinking) and ask Audrey to test this today or tomorrow at the latest.

src/toil_scripts/adam_uberscript/adam_uberscript.py
fnothaft: Ah, fair... That gets a bit tricky...

hannes-ucsc:
Addressed all comments. Pinging @fnothaft for re-review with a focus on `static_adam_preprocessing_dag()` and `remove_file()` to make sure I didn't hose those when resolving conflicts.

fnothaft:
@hannes-ucsc on a call, will review in ~30m.

fnothaft:
LGTM!

hannes-ucsc:
One additional last-minute change in the last commit:

https://github.com/BD2KGenomics/toil-scripts/commit/1bf16122d4c42528d135a4d11a6cdb7b4cda813f#diff-7060afbea50ff85e89fe57836bc27314R564

It makes --zone optional if CGCLOUD_ZONE is defined.

fnothaft:
As an aside @hannes-ucsc, when you said that this was untested in the PR body, I assumed you meant _external Spark cluster autoscaling_, not the modifications to the existing cluster autoscaling code. https://github.com/BD2KGenomics/toil-scripts/commit/df778f8c5e43edf0d22e479188a5aa18e17b8632 contains errors that break the core autoscaling loop. I'm fixing these in #181. That being said, I wouldn't have given an "OK to merge" vote if I had known that the existing autoscaling loop wasn't testet. This is my fault for not asking a question to clarify that, but if code is untested, it should be tagged with `needs work`, no matter how big the patch is, how much code the patch touches, and how much we want to get it merged to avoid conflicts.

hannes-ucsc:
I should have made the risk more clear.

Issue #169, opened 2016-03-13
Rejigger sorting in ADAM/GATK pipeline


fnothaft:
I'm going to move sorting out of alignment and into the `gatk_preprocessing` pipeline. This is because we don't need to sort before running ADAM (and we sort on export), so sorting twice is inefficient.

PR #170, opened 2016-03-13
Add adapter trimming using trimadap via bwakit

### LINKED ISSUE #168: Add adapter trimming to pipeline
There's a bit of confusion from the Harvard email thread as to whether the RAW BAMs have had their adapters trimmed or not. If you look at the hs37d5 BAMs, there are a fair number of hard clipped reads (but these all seem to be supplemental alignments). I'm going to add this to the pipeline (it's baked into BWAKIT, IIRC, so it's a simple add) and I think we should run a sample through with the adapter trimming just to sanity check.

fnothaft:
Adds pipeline option for calling trimadap via bwakit `-a` for trimming Illumina sequencing adapters in-line with alignment. Also, made sorting optional in preparation for #169. Resolves #168. Depends on #153.


hannes-ucsc:
+1

PR #166, opened 2016-03-11
Fixing DAG structure and return value for Spark service cluster (resolves #162)

fnothaft:
Resolves #162.

Doesn't work yet; I'll be offline most of the day tomorrow, but just wanted to push this so that people know what changes I'm making, since we've got lots of things in flight.

PR #171, opened 2016-03-14
Add suffix option to ADAM/GATK pipeline for genome build suffix from überscript

fnothaft:
Testing now. Depends on #166.

### DIFF
--- a/src/toil_scripts/adam_gatk_pipeline/align_and_call.py
+++ b/src/toil_scripts/adam_gatk_pipeline/align_and_call.py
@@ -140,6 +140,11 @@ def build_parser():
     parser.add_argument('-U', '--uuid_manifest', required = True,
                         help = 'Sample UUID.')
 
+    # optionally add suffix
+    parser.add_argument('--dir_suffix',
+                        help = 'Optional suffix to add to output directory names.',
+                        default = '')
+
     # what pipeline are we running
     parser.add_argument('-PR', '--pipeline_to_run',
                         help = "Whether we should run 'adam', 'gatk', or 'both'. Default is 'both'.",
@@ -235,7 +240,8 @@ def sample_loop(job,
                 skip_alignment,
                 skip_preprocessing,
                 autoscale_cluster,
-                sequence_dir):
+                sequence_dir,
+                dir_suffix):
   """
   Loops over the sample_ids (uuids) in the manifest, creating child jobs to process each
   """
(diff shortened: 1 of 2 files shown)


fnothaft:
This is good to merge; I've removed the `needs work` tag.

heuermh:
To clarify, for the purposes of the concordance script, after this is merged there will be the following directories of VCF files to compare
```
s3://{s3_bucket}/analysis{dir_suffix}/{uuid}/
s3://{s3_bucket}/analysis{dir_suffix}/{uuid}/
```
E.g.
```
s3://{s3_bucket}/analysis_GRCh38/{uuid}/
s3://{s3_bucket}/analysis_hg19/{uuid}/
```
I was thinking {GRCh38 vs. hg19} and {GATK vs. ADAM} would be in the file name so I could output concordance results to base filename ```s3://{s3_bucket}/analysis/{a}_vs_{b}```.

I'm fine with the change in this PR, where should the concordance results go then?

fnothaft:
I'd say to add a concordance directory, e.g.:

```
s3://{s3_bucket}/concordance/{a}_vs_{b}
```

Alternatively, I can make a change that would allow the suffix to specify subdirs, i.e., instead of having `analysis_hg19`, we'd have `analysis/hg19`. Then, we could write the GRCh38 ADAM vs. GATK --> `analysis/GRCh38/mySample_concordance_adam_vs_gatk`. We'll also need to write liftover scripts for lifting the GRCh38 VCFs to HG19. I'll open that ticket.

Thoughts? On reflection, I think that `analysis/GRCh38` would be preferable to `analysis_GRCh38`.

hannes-ucsc:
+1 on `analysis/hg19`

And I don't see any validation. If you stick with `_` as the separator, validation should ensure that there are no slashes in the suffix. If you switch to `/`, it should ensure that there is no slash at the end or the beginning of the suffix.

src/toil_scripts/adam_gatk_pipeline/align_and_call.py
```diff
@@ -140,6 +140,11 @@ def build_parser():
     parser.add_argument('-U', '--uuid_manifest', required = True,
                         help = 'Sample UUID.')
 
+    # optionally add suffix
+    parser.add_argument('--dir_suffix',
+                        help = 'Optional suffix to add.',
```
hannes-ucsc: Could probably be a little more substantial.

heuermh:
> +1 on analysis/hg19

Same here

fnothaft:
This is ready for merge, I've removed the `needs work` tag.

Issue #174, opened 2016-03-14
Revive spot support in überscript for Toil cluster and standalone Spark cluster


hannes-ucsc:
It was improved in d1f1adf997d2d7f73ca5a8afeccc6f66dc71e24c and removed by 2d9a0abfba0e6eb21388dc56cfdfe07f056649a9.

hannes-ucsc:
Dupe of #256

Issue #126, opened 2016-03-01
Spark/HDFS leader can die, causing our cluster to discover the true depths of sadness


fnothaft:
See #123. Essentially, since we bid spot instances for all of the instances in our cluster, the Spark master and HDFS namenode run on a spot instance. If and when this instance is revoked, our cluster achieves a deepness of despair that is truly bleak. Our worker nodes fail, but cannot be recovered since our Spark/HDFS workers and Spark jobs rely on knowing the master IP via parameters, and we cannot meaningfully restart them (pointed at a new master) without modifying the jobstore.

To fix this, we have worked out a modest proposal:

* @hannes-ucsc will add hooks to the überscript that identify when a cluster has lost a node, and kills all of the nodes in a cluster. This is crude, but suffices.
* @hannes-ucsc will update cgcloud so that we can tag instances (e.g., the master), which makes it easy to dynamically find the master IP.
* @fnothaft will restructure the `adam_pipeline` to resolve some other issues (job decomposition) that make it impossible to reliably restart the ADAM jobs due to lack of input/output checkpointing

Or, just run on-demand instances.

fnothaft:
- [ ] @hannes-ucsc https://github.com/BD2KGenomics/toil/issues/706
- [ ] @fnothaft #134

heuermh:
![Negative Comic Book Guy](http://i.imgur.com/M4g9uAb.jpg)

jvivian:
Will be resolved by BD2KGenomics/toil#786

PR #181, opened 2016-03-15
Fix cluster shrinking errors (resolves #172)

### LINKED ISSUE #172: Cluster shrinker is... bloodthirsty
I've been having an enjoyable experience recently, as the cluster shrinking code from the überscript has gone all Jason Voorhees and my toil-workers are being cut down in the streets, in front of my own, persistently weepy eyes.

There are two problems that I am aware of:

* Our [CPU utilization threshold](https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/adam_uberscript/adam_uberscript.py#L326) is unrealistically high (50%). Some of our pipeline stages are IO bound (MarkDups pegs I/O while keeping CPU ~25%, IIRC), and also we have nodes chilling around without doing a ton of work (i.e., the Spark masters). This is easy to fix.
* We are killing nodes when we are at our requested cluster size. There's a [conditional that checks this](https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/adam_uberscript/adam_uberscript.py#L404), but for some reason, this isn't firing properly. I've added logging to debug this. This manifests as:
 * We bump the node count.
 * We don't decrease the node count.
 * ...
 * The autoscaler kills off the idle nodes, and immediately restarts them to bring the cluster up to desired size.
 * PROFIT! (LOSS?)

fnothaft:
* Fixed off by one error in `get_desired_cluster_size` and removed one-off fix
  in `grow_cluster`.
* Added while loop to spin and check for instance termination.
* Fixed synchronization bug where cluster size check happened outside of lock
  in `grow_cluster`.
* Added logging around per-sample instance requests.

### DIFF
--- a/src/toil_scripts/adam_uberscript/adam_uberscript.py
+++ b/src/toil_scripts/adam_uberscript/adam_uberscript.py
@@ -38,8 +38,8 @@ metric_initial_wait_period_in_seconds = 0
 metric_collection_interval_in_seconds = 3600
 metric_start_time_margin = 1800
 
-scaling_initial_wait_period_in_seconds = 300
-cluster_scaling_interval_in_seconds = 300
+scaling_initial_wait_period_in_seconds = 60
+cluster_scaling_interval_in_seconds = 150
 
 # Protects against concurrent changes to the size of the Toil cluster, by both the metric thread terminating idle
 # nodes and the cluster-growing thread adding nodes.
@@ -234,7 +234,14 @@ def parse_cgcloud_list_output(output):
 
 def get_desired_cluster_size(domain):
     nodes_per_sample = Samples.load(domain)
-    return sum(map(lambda x: x[1], nodes_per_sample.samples.values()))
+
+    sample_request = sum(map(lambda x: x[1].nodes, nodes_per_sample.value.iteritems()))
+    log.info("Samples are currently requesting %d nodes: %s",
+             sample_request,
+             nodes_per_sample.value)
+
+    # add one because there needs to be a single communal worker, else deadlock is upon us
+    return sample_request + 1
 
 
 def update_cluster_size(domain, n):
@@ -286,7 +293,7 @@ def grow_cluster(num_nodes, instance_type, cluster_name, cluster_type='toil', *o
 def manage_metrics_and_cluster_scaling(params):
     conn = boto.sdb.connect_to_region(region_of_zone(params.zone))
     dom = conn.get_domain('{0}--files'.format(params.jobstore))
-    grow_cluster_thread = threading.Thread(target=manage_toil_cluster, args=(params, conn, dom))
+    grow_cluster_thread = threading.Thread(target=manage_toil_cluster, args=(params, dom))
     metric_collection_thread = threading.Thread(target=collect_realtime_metrics, args=(params,))
     grow_cluster_thread.start()
     metric_collection_thread.start()
@@ -323,10 +330,13 @@ def manage_toil_cluster(params, domain):
     time.sleep(scaling_initial_wait_period_in_seconds)
     while True:
         with throttle(cluster_scaling_interval_in_seconds):
-            cluster_size = len(list_nodes(params.cluster_name))
-            desired_cluster_size = get_desired_cluster_size(domain) + 1
-            if cluster_size < desired_cluster_size:
-                with cluster_size_lock:
+            with cluster_size_lock:
+                cluster_size = len(list_nodes(params.cluster_name))
+                desired_cluster_size = get_desired_cluster_size(domain)
+                if cluster_size < desired_cluster_size:
+                    log.info("Cluster size (%d) is smaller than requested (%d).",
+                             cluster_size,
+                             desired_cluster_size)
                     num_nodes = desired_cluster_size - cluster_size
                     grow_cluster(num_nodes, params.instance_type, params.cluster_name, *role_options(params))
                     update_cluster_size(domain, desired_cluster_size)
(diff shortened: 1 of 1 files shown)


src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -465,10 +475,29 @@ def collect_realtime_metrics(params, threshold=0.5):
...
+                                         cluster_size,
+                                         desired_cluster_size,
+                                         instance_id)
+
+                                terminated_instances = conn.terminate_instances(instance_ids=[instance_id])
+                                assert len(terminated_instances) == 1, 'Terminating instance %s failed.' % instance_id
+
+                                # it is not clear as to whether boto's terminate call is blocking
+                                # so let's spin for a bit, because if we don't block on the
+                                # instance termination, we get a FUBAR synchronization bug
```
hannes-ucsc: None of the EC2 operations that trigger state changes of resources like instances, volumes, snapshots, spot requests are blocking.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -465,10 +475,29 @@ def collect_realtime_metrics(params, threshold=0.5):
...
+                                cur_size = get_cluster_size(params.cluster_name)
```
hannes-ucsc: get_cluster_size() is rather expensive.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -465,10 +475,29 @@ def collect_realtime_metrics(params, threshold=0.5):
...
+                                while cur_size >= cluster_size:
```
hannes-ucsc: The cluster size could go down for various reasons, e.g. spot termination of other instances.

https://github.com/BD2KGenomics/cgcloud/pull/129 is currently being tested. It will allow you to synchronously terminate an instance by ID.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -388,7 +398,7 @@ def manage_standalone_spark_cluster(params, domain):
         semaphore = Semaphore.load(domain, name=standalone_spark_semaphore_name)
 
 
-def collect_realtime_metrics(params, threshold=0.5):
+def collect_realtime_metrics(params, threshold=0.025):
```
hannes-ucsc: I'm pretty sure .5 is 0.5% not 50%. @jvivian can you confirm?

fnothaft: Yeah, after reviewing the logs from my latest run, looks like 0.5% not 50%. I'll remove that change.

src/toil_scripts/adam_uberscript/adam_uberscript.py
fnothaft: Agreed. However, I don't think that the überscript properly handles spot termination right now anyways (see `ClusterSize` class), and I'm only running on in-demand clusters due to #126, and anyways we don't _actually_ support spot instances right now anyways (see #174).

src/toil_scripts/adam_uberscript/adam_uberscript.py
fnothaft: Sure, but I don't think you can avoid the call here without moving to the cgcloud call. How close are we to https://github.com/BD2KGenomics/cgcloud/pull/129 being merged?

fnothaft: If we're talking https://github.com/BD2KGenomics/cgcloud/pull/129 merges today, then let's wait on that and I'll update this PR and we can remove a lot of smelly code.

fnothaft:
@hannes-ucsc updated to use BD2KGenomics/cgcloud#129 and to resolve the % issue. I am testing now and will report back.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -465,10 +475,30 @@ def collect_realtime_metrics(params, threshold=0.5):
...
+                                         'Terminating idle instance %s.',
+                                         cluster_size,
+                                         desired_cluster_size,
+                                         instance_id)
+                                
+                                cmd = ['cgcloud',
+                                       'terminate',
+                                       '--instance-id', instance_id,
+                                       '--cluster-name', params.cluster_name,
+                                       '--quick',
```
hannes-ucsc: Don't want to use `--quick` here, I think. It does not block.

src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -465,10 +475,30 @@ def collect_realtime_metrics(params, threshold=0.5):
...
+                                       'toil']
+
+                                try:
+                                    check_call(cmd)
+                                    log.info("Successfully terminated instance via %s.",
+                                             " ".join(cmd))
```
hannes-ucsc: I usually do `log.info("Successfully terminated instance via %r.", cmd)` because that's not lossy in case there are spaces in any of the elements of cmd.

fnothaft:
Can has merge plz?

hannes-ucsc:
@fnothaft I was trying to be polite and phrased it as a nice to have but I still think the %r vs %s thing for commands is important. Next time I'll be more assertive ;-)

fnothaft:
> @fnothaft I was trying to be polite and phrased it as a nice to have but I still think the %r vs %s thing for logging commands is important. Next time I'll be more assertive ;-)

Ah, sorry, I missed that. I'll add a follow-on tomorrow.

PR #187, opened 2016-03-16
Miscellaneous adam_gatk_pipeline fixes (resolves #169 et al)

### LINKED ISSUE #169: Rejigger sorting in ADAM/GATK pipeline

### LINKED ISSUE #176: Clean up GATK preprocessing pipeline
Misc issues:

* We sort, index, resort, reindex.
* Disk settings aren't given properly in `adam_gatk_pipeline`, so we OOD.

### LINKED ISSUE #177: BQSR result size is too large
Bigger than `spark.driver.maxResultSize`:

```
WARNING:toil.leader:7337aec6-1c64-41ff-8684-46c35d624c08:       org.apache.spark.SparkException: Job aborted due to stage failure: Total size of serialized results of 247 tasks (1024.4 MB) is bigger than spark.driver.maxResultSize (1024.0 MB)
```

Almost all of the time it is by 5 damned MB.

### LINKED ISSUE #178: Read @RG data in from manifest
We should read the record group data in from our manifest file when running the ADAM/GATK pipeline. We have this data from the SGDD raw alignments, so if we're going to the trouble of generating a pipeline and BAM/VCF files that people will hopefully use someday, we might as well do this part correctly too.

### LINKED ISSUE #184: multiprocessing called from toil leader in adam_gatk_pipeline
E.g., https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/adam_gatk_pipeline/align_and_call.py#L518

### LINKED ISSUE #185: Why—oh god why—are we passing ints as strings?
While fixing #184, I found that we were calling `str(multiprocessing.cpu_count())`, which I thought was bizarre. It turns out that this is an optimization, because we wind up passing this `cpu_count` to the GATK as a core count, and since that call is going through `docker_call`/`check_call`, etc., it needs to be a string, and so on.

But, upon further reflection here and in #176, it turns out that this is sort of why we're getting OOD errors in the GATK pipelines. Since the cores argument is never set on any of the toil jobs in the GATK germline/preprocessing pipelines, the parameter defaults to 1 core. This leads to the occasional instance where multiple GATKs get put on a single node, which leads to good times, etc. You don't want to blanket set the core count (or disk requirements) because there are various parts of the pipeline that are single threaded, only operate on a reference file, etc.

So, while passing ints as strings is not always wrong, in this case, it was hiding a job resource request bug which led to tons of funny business.

### LINKED ISSUE #188: Picard MarkDups may fail, but Toil lives to die another day
Haven't root caused the exact reason that MarkDups is failing, nor have I root caused why the job isn't failing, but we see MarkDups fail (writes an error to the log, doesn't write output), and then toil continues and runs INDEL realignment, which does fail (because it can't find it's input file). I'll fix this and tack it on to #187.

### LINKED ISSUE #189: GATK wants karyotypic order for HG19
Well you know, your reference files might be in karyotypic order and it'd be a damn shame if your reads weren't in karyotypic order because that would be a completely unrecoverable error now wouldn't it be. I mean, since all of our inputs are indexed, there's no possible way that we could use the index to read  the BAM file in random order.

Also, since we validate the order of ref sequences listed in all inputs, there's no possible way that we could see our reference, known-sites VCFs, and reads have the same ref sequence order, and skip the karyotypic order.

Anyways, word on the street is that you can disable this with `--unsafe`. Going to add this to #187 as well.

Possibly the best part of this is that someone hardcoded both karyotypic order and code to recognize the HG19 reference build into the GATK.

### LINKED ISSUE #191: Picard MarkDuplicates denies the existence of the SAM 0x800 flag
I give up: https://sourceforge.net/p/samtools/mailman/message/32800679/. Supposedly this was fixed in Picard 1.96? That would make sense, if we weren't running:

```
$ docker run quay.io/ucsc_cgl/picardtools:1.95--dd5ac549b95eb3e5d166a5e310417ef13651994e --version
picard-tools
Copyright 2010 The Broad Institute
```

?????????
\<pained facial expression\>
More sleuthing:

```
$  apt-get changelog picard-tools
picard-tools (1.105-1) unstable; urgency=low

  * New upstream release 1.105
  * Conforms with Policy 3.9.5.

 -- Olivier Sallou <osallou@debian.org>  Tue, 24 Dec 2013 11:16:18 +0100
```

Which checks out when cross referenced with the Picard releases page.

[Anyways, as far as I can tell, the fix isn't to—you know—fix Picard](https://www.biostars.org/p/75693/), but rather, to run `bwa mem` with the `-M` flag ("Mark shorter split hits as secondary (for Picard compatibility).") because the correct way to fix a trivial bug is to introduce intentionally incorrect behavior into another tool.

[\<another pained expression\>](https://www.biostars.org/p/176814/)

Anyways, it's not clear if we can pass the `-M` flag via bwakit. I'm not much of a Perl person, but AFAIK, there's no way to [add to the bwa mem options here](https://github.com/lh3/bwa/blob/master/bwakit/run-bwamem#L146).

So, TBD.

### LINKED ISSUE #197: Bump to GATK 3.5
I'm running into a [HTSJDK-related issue in the GATK](http://gatkforums.broadinstitute.org/gatk/discussion/6408/error-haplotypecaller) that is supposedly resolved in GATK 3.5.

### LINKED ISSUE #198: Only use S3 URLs in adam_überscript
The HG19 ACLs in `s3://cgl-pipeline-inputs/` are b0rk3d. Using `s3://` URLs seems to resolve this issue.

fnothaft:
Depends on #171. Still testing it; I get a pickling error somewhere in the GATK.

Resolves #169, resolves #176, resolves #177, resolves #178, resolves #184, resolves #185, resolves #188, resolves #189, resolves #191, resolves #197, resolves #198.

Alternate working title: "The God PR"


fnothaft:
Pushed an update that resolves the pickling issues, but now Picard Markdups is throwing an internal error. Looking into that...

hannes-ucsc:
Sorry for the long title. That's just what Waffle and GH want.

fnothaft:
@hannes-ucsc not sure about Waffle, but GitHub will close an issue when a commit containing "Resolves #number" or "Closes #number" is pushed. It doesn't have to be in the PR body/title.

fnothaft:
Added a fix for #191, and updated the title. Testing this now...

hannes-ucsc:
Waffle seems to be happy with the issue linked in the body.

fnothaft:
If anyone would like to review this, review it now, because it looks like we can remove the `needs work` tag as soon as https://github.com/BD2KGenomics/cgl-docker-lib/pull/99 merges and pushes to Quay.

fnothaft:
TODO:

* Fix #204 
* Correct BAM index paths for INDEL realigner and BQSR output files

fnothaft:
This is rebased. I don't have a setup for testing this right now (specifically, [the newest commit that resolves](https://github.com/BD2KGenomics/toil-scripts/pull/187/commits/7c4e64f8cf8c6640fce7e39cad6f828f9d076c6c) #204, and a mod to [the indexing commit](https://github.com/BD2KGenomics/toil-scripts/pull/187/commits/a82db1639b0005daf1f73c4ac7ca91405db3613e), all else is tested). How do we want to proceed? Do we want to remove the `needs work` tag and move forward? Do we want to halt until we have the credit situation resolved?

fnothaft:
Ping @hannes-ucsc @jvivian for merge. This is blocking @almussel.

Issue #111, opened 2016-02-23
work_dir should be optional in docker_call


fnothaft:
We pass CWD when using docker_call with Spark jobs, which seems a bit silly.

heuermh:
This one
https://github.com/BD2KGenomics/toil-scripts/blob/master/src/toil_scripts/batch_alignment/bwa_alignment.py#L178

or should ```docker_call``` be moving to some util class?

hannes-ucsc:
I've told at least three people to put their stuff into `toil_scripts/__init__.py` but I am worried that this will create unnecessary merge conflicts. Shall we start putting shared code into `toil_scripts/utils`, e.g  `toil_scripts/utils/docker.py` for docker related stuff?.

fnothaft:
> but I am worried that this will create unnecessary merge conflicts.

As long as things are independent, they should be easy to resolve. I wouldn't worry too much, but:

> Shall we start putting shared code into toil_scripts/utils, e.g `toil_scripts/utils/docker.py` for docker related stuff?

Also SGTM.

jvivian:
Ya, i have a PR I'm working on right now for a pipeline for doing alternative splicing analysis, and I've been exporting all the "general" functions to `__init__`.     I just made comments that separate the AWS , docker, job, and general functions. 

> As long as things are independent, they should be easy to resolve. I wouldn't worry too much, but:

I agree in some sense, but I've also been making modifications to a lot of these to make them simple and general purpose. I've seen a few that  get passed dictionaries where _specific_ key names are required, etc.  

In parallel, I'm also trying to write unit tests for all the functions I'm moving out.  Maybe I can coordinate with Hannes about breaking my PR up into a couple different ones, or what the best approach would be.

fnothaft:
> I'd favor smaller PRs since this is a busy project ATM.

+1, small PRs were my assumption when I said:

> As long as things are independent, they should be easy to resolve.

jvivian:
Resolved by common library refactoring

PR #190, opened 2016-03-16
Add common library (resolves #112, resolves #227)

### LINKED ISSUE #112: Dedup mess of duplicated functions in GATK pipelines
E.g., `docker_call`, `download_from_url`, other functions that just share a name, etc.

Many of these should be pulled out into a dedicated utils module, or something of that sort.

### LINKED ISSUE #227: Pin versions of pip-installed dependencies in jenkins.sh
… to prevent updates to a dependency from breaking our build.

jvivian:
Checked that tests pass on multiple machines.


jvivian:
@hannes-ucsc Ready for review

src/toil_scripts/lib.py
```diff
@@ -0,0 +1,276 @@
...
+        download_from_s3_url(file_path, url)
+    else:
+        try:
+            subprocess.check_call(['curl', '-fs', '--retry', '5', '--create-dir', url, '-o', file_path])
+        except OSError:
+            raise RuntimeError('Failed to find "curl". Install via "apt-get install curl"')
+    assert os.path.exists(file_path)
+
+
+# FIXME -- Remove SUDO support to Docker Call?
```
hannes-ucsc: @fnothaft thinks we should leave this in. Let's remove teh FIMXE

src/toil_scripts/lib.py
```diff
@@ -0,0 +1,276 @@
...
+        else:
+            exit_code = subprocess.check_call(base_docker_call + [tool] + parameters)
+    except subprocess.CalledProcessError:
+        raise RuntimeError('docker command returned a non-zero exit status: {}'.format(parameters))
+    except OSError:
+        raise RuntimeError('docker not found on system. Install on all nodes.')
+    return exit_code
+
+
+# AWS
```
hannes-ucsc: remove

src/toil_scripts/lib.py
```diff
@@ -0,0 +1,276 @@
...
+    else:
+        s3am_with_retry(num_cores, 'file://{}'.format(fpath), s3_dir)
+
+
+# Encryption
+def generate_unique_key(master_key_path, url):
+    """
+
+    master_key_path: str    Path to the BD2K Master Key (for S3 Encryption)
+    url: str                S3 URL (e.g. https://s3-us-west-2.amazonaws.com/bucket/file.txt)
```
hannes-ucsc: Should we switch to s3:// URLs before we merge this?

src/toil_scripts/test_lib.py
```diff
@@ -0,0 +1,106 @@
+"""
```
hannes-ucsc: The common convention is to put tests into a `test` subpackage.

src/toil_scripts/lib.py
```diff
@@ -0,0 +1,276 @@
...
+
+    Returns: str            32-byte unique key generated for that URL
+    """
+    with open(master_key_path, 'r') as f:
+        master_key = f.read()
+    assert len(master_key) == 32, 'Invalid Key! Must be 32 characters. ' \
+                                  'Key: {}, Length: {}'.format(master_key, len(master_key))
+    new_key = hashlib.sha256(master_key + url).digest()
+    assert len(new_key) == 32, 'New key is invalid and is not 32 characters: {}'.format(new_key)
+    return new_key
```
hannes-ucsc: missing NL

jvivian:
@hannes-ucsc Addressed all issues besides changing encryption to s3:// prefix.  Implementing this fix right away would make most of our input data unretrievable. 

Should I assign myself a ticket to automate the rencryption / deletion of old files given a bucket?  Once I've written the script we'll have to apply it to the cgl-driver-projects-encrypted bucket and the four recompute dataset buckets on firecloud.  

There's also the issue of this conflicting with the releases/1.0 branch, but should be manageable.

src/toil_scripts/test_lib.py
hannes-ucsc: Most importantly, we should run these tests on Jenkins.

src/toil_scripts/test/__init__.py
```diff
@@ -0,0 +1 @@
+__author__ = 'Jvivian'
```
hannes-ucsc: no author tags

src/toil_scripts/test/test_lib.py
```diff
@@ -0,0 +1,106 @@
+"""
+Computational Genomics Lab
+Author: John Vivian
```
hannes-ucsc: no author tags

hannes-ucsc:
Lets create a jenkins.sh that runs the tests.

hannes-ucsc:
> Should I assign myself a ticket to automate the rencryption / deletion of old files given a bucket? Once I've written the script we'll have to apply it to the cgl-driver-projects-encrypted bucket and the four recompute dataset buckets on firecloud.

Ticket, yes. Start work, probably not yet. The latest s3am has caching of temp creds so we should be able to insanely oversubscribe a box with like 32 s3am instance each one doing a bucket to bucket copy. Maybe this should be a Toil script. In fact, I think it should.

> There's also the issue of this conflicting with the releases/1.0 branch, but should be manageable.

Sorry, what conflict are you referring to?

hannes-ucsc:
Can we switch to the following layout:

```
src/toil_scripts/lib/urls.py (download_...)
src/toil_scripts/lib/files.py (mkdir_p, ...)
src/toil_scripts/lib/programs.py (which, docker_call ...)
src/toil_scripts/lib/__init__.py (other stuff, except job functions due to bug in Toil)
src/toil_scripts/lib/test/test_urls.py
```

I know it looks like overkill right now but will be more sustainable in the future.

I think job functions should always end in `_job`.

fnothaft:
+1 RE: layout change, job function naming semantics (document in README?)

jvivian:
> Sorry, what conflict are you referring to?

All the scripts on *releases/1.0* pull encrypted data using the current encryption method.  We'll just have to make sure that if we reencrypt all the data that all the scripts (which _don't_ pull from lib) have their generate_function_keys refactored.

jvivian:
> +1 RE: layout change, job function naming semantics
> Can we switch to the following layout:

Addressed.

Sourcetree isn't letting me squash these two commits, tried several different things but when I try rebase interactively on the parent the window never opens (it works on other commits). 

¯\_(ツ)_/¯

jvivian:
lib.urls has been collapsed to have single functions for `download_url` and `s3am_upload`.   There is a DRY `_job` version of each of those functions as well. 

I added `consolidate_tarballs_job()` to **test_files.py**

Unittests have been written for all functions — `s3am_upload` and `download_url` with encryption are  tested in a single unittest (**test_url.py**), which generates a unique encryption key, uploads a test file with encryption, downloads the encrypted file, then deletes the key from S3.  S3 destination is the UCSC-BD2K account, so Jenkins will need ~/.boto access to ucsc-bd2k to run these. 

Tests take 10 seconds to run on my machine (TOIL==3.1.4), but will take a little longer the first time as I test docker_call with the samtools image. 

@hannes-ucsc @fnothaft feedback appreciated

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,49 @@
...
+    Makes subprocess call of a command to a docker container.
+
+
+    tool_parameters: list   An array of the parameters to be passed to the tool
+    tool: str               Name of the Docker image to be used (e.g. quay.io/ucsc_cgl/samtools)
+    java_opts: str          Optional commands to pass to a java jar execution. (e.g. '-Xmx15G')
+    outfile: file           Filehandle that stderr will be passed to
+    sudo: bool              If the user wants the docker command executed as sudo
+    """
+    base_docker_call = 'docker run --log-driver=none --rm -v {}:/data'.format(work_dir).split()
```
hannes-ucsc: use list literal not "...".split()

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,49 @@
...
+    if sudo:
+        base_docker_call = ['sudo'] + base_docker_call
```
hannes-ucsc: use list.insert()

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,49 @@
...
+    if java_opts:
+        base_docker_call = base_docker_call + ['-e', 'JAVA_OPTS={}'.format(java_opts)]
```
hannes-ucsc: Make this generic.

docker_call( ..., env=dict(JAVA_OPTS="-Xmx")

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,49 @@
...
+    try:
+        if outfile:
+            exit_code = subprocess.check_call(base_docker_call + [tool] + parameters, stdout=outfile)
```
hannes-ucsc: check_call() returns None

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,49 @@
...
+        else:
+            exit_code = subprocess.check_call(base_docker_call + [tool] + parameters)
+    except subprocess.CalledProcessError:
+        raise RuntimeError('docker command returned a non-zero exit status: {}'.format(parameters))
```
hannes-ucsc: Why?

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,49 @@
...
+    except OSError:
```
hannes-ucsc: Too unspecific

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,49 @@
...
+        raise RuntimeError('docker not found on system. Install on all nodes.')
```
hannes-ucsc: Again, why? Doesn't the OSError convey exactly the same info?

src/toil_scripts/lib/test/test_files.py
```diff
@@ -0,0 +1,58 @@
+import os
+import tarfile
+from toil.job import Job
+
+
+def test_mkdirp(tmpdir):
```
hannes-ucsc: test_mkdir_p

src/toil_scripts/lib/test/test_files.py
```diff
@@ -0,0 +1,58 @@
...
+    import os
+    from toil_scripts.lib.files import mkdir_p
+    dir_path = os.path.join(str(tmpdir), 'test')
+    # Exists -- shouldn't throw error
```
hannes-ucsc: prove that it exists

src/toil_scripts/lib/test/test_files.py
```diff
@@ -0,0 +1,58 @@
+import os
+import tarfile
+from toil.job import Job
+
+
+def test_mkdirp(tmpdir):
+    import os
+    from toil_scripts.lib.files import mkdir_p
+    dir_path = os.path.join(str(tmpdir), 'test')
```
hannes-ucsc: isn't tmpdir a string already?

src/toil_scripts/lib/test/test_files.py
```diff
@@ -0,0 +1,58 @@
...
+    # Exists -- shouldn't throw error
+    mkdir_p(str(tmpdir))
+    mkdir_p(dir_path)
```
hannes-ucsc: I guess we don't need to test the automatic creation of the parent directories.

src/toil_scripts/lib/test/test_files.py
```diff
@@ -0,0 +1,58 @@
...
+    assert os.path.isdir(dir_path)
+
+
+def test_tarball_files(tmpdir):
```
hannes-ucsc: Note to self: I stopped here.

hannes-ucsc:
Still need jenkins.sh so we can run the tests on CI.

Many files have newlines missing at end of file.

hannes-ucsc:
PR and commit title are wrong. resolves not resolved

hannes-ucsc:
And can you switch the docstrings to the Sphinx syntax?

```
:param str foo: name of your bar
```

src/toil_scripts/lib/programs.py
jvivian: screen shot 2016-03-19 at 8 06 28 am

src/toil_scripts/lib/test/test_files.py
jvivian: `print type(tmpdir)`

class 'py._path.local.LocalPath'

jvivian:
@hannes-ucsc Thanks for the great suggestions!

- Fixed docker_call: added arbitrary number of ENV variables. Removed unnecessary exception catches.
- Fixed test_mkdir_p: name change, added assertion that tmpdir exists
- Fixed all docstrings
- Fixed PR and commit title

> jenkins.sh

I looked at the Toil one as an example — for this repo would it just contain `py.test`?

src/toil_scripts/lib/programs.py
hannes-ucsc: Fine. So the return value is always 0 instead of always None. What's the difference? It still is pointless to return that, don't you agree?

hannes-ucsc:
>> Still need jenkins.sh so we can run the tests on CI.

> I looked at the Toil one as an example — for this repo would it just contain `py.test`?

I'm not sure. Whatever it takes to run the tests in a way that fails the build if a test fails. And we need this:

https://pytest.org/latest/usage.html#creating-junitxml-format-files

jvivian:
@hannes-ucsc — Addressed line comments.  Maybe tomorrow we can setup the Jenkins config?  Since it only runs `py.test`, it seems like it should only run if someone modifies the toil_scripts/lib directory until we get test samples for all the pipelines.

jenkins.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
```
hannes-ucsc: No need for shebang. This file must be sourced and having the shebang would suggest otherwise.

hannes-ucsc:
https://github.com/BD2KGenomics/toil-scripts/pull/190#issuecomment-198663118

jvivian:
> #190 (comment) 

Crap, didn't realize Sphinx structure was different than PyCharm's automatically generated one.

hannes-ucsc:
You can tell PyCharm to generate Sphinx docstrings. I recommend you do that.

jenkins.sh
```diff
@@ -0,0 +1 @@
+py.test --junitxml=results.xml
```
hannes-ucsc: Sorry, didn't see that until now. Can you change this to `test-report.xml`?

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Build fails on Jenkins because py.test is not installed. Installation of py.test (in a virtualenv !!!) should be part of jenkins.sh.

jvivian:
Jenkins, test this please.

jvivian:
@hannes-ucsc — Can you trigger a build please? I don't have privileges (here or on jenkins.cgcloud.info). Thanks!

hannes-ucsc:
I tweaked the build config. Can you ask Jenkins to build this? Be sure to be polite:

https://github.com/BD2KGenomics/toil-scripts/pull/190#issuecomment-199497609

hannes-ucsc:
My previous comment triggered a build. But you should be able to trigger one too.

jvivian:
@hannes-ucsc — Is Docker installed on the jenkins-box?  I might also have to specify `py.test src/toil_scripts/lib` in the jenkins.sh if it's running tests on site-packages

```
platform linux2 -- Python 2.7.6, pytest-2.9.1, py-1.4.31, pluggy-0.3.1
rootdir: /var/lib/jenkins/workspace/toil-scripts, inifile: 
collected 60 items / 6 errors

src/toil_scripts/lib/test/test_lib.py ..
src/toil_scripts/lib/test/test_programs.py .F
venv/lib/python2.7/site-packages/wheel/test/test_basic.py ......F..
venv/lib/python2.7/site-packages/wheel/test/test_install.py ..
venv/lib/python2.7/site-packages/wheel/test/test_keys.py ..
venv/lib/python2.7/site-packages/wheel/test/test_paths.py .
venv/lib/python2.7/site-packages/wheel/test/test_ranking.py .
venv/lib/python2.7/site-packages/wheel/test/test_signatures.py ...
venv/lib/python2.7/site-packages/wheel/test/test_tagopt.py .......
venv/lib/python2.7/site-packages/wheel/test/test_tool.py F
venv/lib/python2.7/site-packages/wheel/test/test_wheelfile.py ..
venv/local/lib/python2.7/site-packages/wheel/test/test_basic.py E..FF.F..
venv/local/lib/python2.7/site-packages/wheel/test/test_install.py ..
venv/local/lib/python2.7/site-packages/wheel/test/test_keys.py ..
venv/local/lib/python2.7/site-packages/wheel/test/test_paths.py .
venv/local/lib/python2.7/site-packages/wheel/test/test_ranking.py .
venv/local/lib/python2.7/site-packages/wheel/test/test_signatures.py ...
venv/local/lib/python2.7/site-packages/wheel/test/test_tagopt.py EEEEEEE
venv/local/lib/python2.7/site-packages/wheel/test/test_tool.py F
venv/local/lib/python2.7/site-packages/wheel/test/test_wheelfile.py ..
```

hannes-ucsc:
Until now, the project wasn't restricted to a particular type of slave. I just fixed that such it now requests a toil-jenkins-slave which has Docker on it. We could also use docker-jenkins-slave which we use for cgl-docker-lib.

hannes-ucsc:
Ignore this.

jenkins.sh
```diff
@@ -0,0 +1,4 @@
+virtualenv venv
+. venv/bin/activate
+pip2.7 install pytest toil boto s3am
+py.test src/toil_scripts/lib --junitxml=test-report.xml
```
hannes-ucsc: `src` instead of `src/toil_scripts/lib`

jenkins.sh
```diff
@@ -0,0 +1,4 @@
+virtualenv venv
+. venv/bin/activate
+pip2.7 install pytest toil boto s3am
```
hannes-ucsc: You shouldn't mix multiple applications in a single virtualenv. The reason we use virtualenv's is to isolate applications and their dependencies from each other. In fact, the most recent build error

http://jenkins.cgcloud.info/job/toil-scripts/11/testReport/test/test_urls/test_upload_and_download_with_encryption/

is a result of mixing s3am and Toil in a venv. They each depend on a different version of bd2k-python-lib and one wins. In this case the older one from Toil wins and that breaks s3am.

jvivian:
Added `sample_batcher_job` to jobs.py with accompanying test in test/test_jobs.py

src/toil_scripts/lib/__init__.py
```diff
@@ -0,0 +1,31 @@
+def flatten(x):
+    """
+    Flattens a nested array into a single list
```
hannes-ucsc: Add `FIXME: replace with bd2k.util.iterables.flatten`

src/toil_scripts/lib/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+        if hasattr(el, "__iter__") and not isinstance(el, basestring):
+            result.extend(flatten(el))
+        else:
+            result.append(el)
+    return result
+
+
+def partitions(l, partition_size):
+    """
+    >>> list(partitions([], 10))
```
hannes-ucsc: Can you make sure that the doctests are run by py.test

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,81 @@
+from contextlib import closing
+import os
+import errno
+import tarfile
+import shutil
+
+
+def mkdir_p(path):
+    """
+    It is Easier to Ask for Forgiveness than Permission
```
hannes-ucsc: Add `FIXME: replace with bd2k.util.files.mkdir_p`

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+def tarball_files(tar_name, work_dir='.', fpaths=None, uuid=None):
+    """
+    Creates a tarball from a group of files
+
+    :param str tar_name: Name of tarball
+    :param str work_dir: Current working directory
+    :param list fpaths: File paths to include in the tarball
+    :param str uuid: UUID to stamp files with
```
hannes-ucsc: `:param str prefix: optional prefix for the names of the files in the archive`

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,81 @@
...
+    """
+    if fpaths is None:
+        fpaths = []
+    with tarfile.open(os.path.join(work_dir, tar_name), 'w:gz') as f_out:
+        for fpath in fpaths:
+            if uuid:
+                f_out.add(fpath, arcname=uuid + '.' + os.path.basename(fpath))
```
hannes-ucsc: Remove '.' then people can chose whatever prefix they want.

There too much duplicated code in the if and else branches.

hannes-ucsc: I meant "separator" not "prefix".

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,81 @@
...
+            else:
+                f_out.add(fpath, arcname=os.path.basename(fpath))
+
+
+def move_to_output_dir(output_dir, filepaths=None):
+    """
+    Moves files from the working directory to the output directory.
+
+    :param str output_dir: Output directory
+    :param list filepaths: Filepaths to move
```
hannes-ucsc: `list[str]` and `file paths` not `filepaths`

hannes-ucsc: Also shorten function name to `move_files` swap arguments. A common convention for argument orderin is usually f(source, destination), just like the `cp` and `mv` command line utilities.

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,81 @@
...
+        dest = os.path.join(output_dir, os.path.basename(fpath))
+        shutil.move(fpath, dest)
+
+
+def consolidate_tarballs_job(job, **kwargs):
+    """
+    Combine the contents of separate tarballs into one.
+    Subdirs within the tarball will be named the keys in **kwargs.
+
+    :param dict kwargs: Key word arguments of the form: file_name_prefix=FileStoreID
```
hannes-ucsc: `dict[str,str]`

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,81 @@
...
+    :return str: FileStoreID
+    """
+    work_dir = job.fileStore.getLocalTempDir()
+    # Retrieve output file paths to consolidate
+    tar_paths = []
+    for fname in kwargs:
```
hannes-ucsc: Replace with `for fname, fileStoreId in kwargs.iteritems():`

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,81 @@
...
+    :param list filepaths: Filepaths to move
+    """
+    if filepaths is None:
+        filepaths = []
+    for fpath in filepaths:
+        dest = os.path.join(output_dir, os.path.basename(fpath))
+        shutil.move(fpath, dest)
+
+
+def consolidate_tarballs_job(job, **kwargs):
```
hannes-ucsc: Rename `kwargs` to `fname_to_id`

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,81 @@
...
+    out_tar = os.path.join(work_dir, output_name)
+    # Consolidate separate tarballs into one
+    with tarfile.open(os.path.join(work_dir, out_tar), 'w:gz') as f_out:
+        for tar, fname in tar_paths:
+            with tarfile.open(tar, 'r') as f_in:
+                for tarinfo in f_in:
+                    with closing(f_in.extractfile(tarinfo)) as f_in_file:
+                        tarinfo.name = os.path.join(output_name, fname, os.path.basename(tarinfo.name))
+                        f_out.addfile(tarinfo, fileobj=f_in_file)
+    return job.fileStore.writeGlobalFile(out_tar)
```
hannes-ucsc: I wonder if the generality of this method would be increased if you removed the call to wGF().

src/toil_scripts/lib/jobs.py
```diff
@@ -0,0 +1,20 @@
...
+def sample_batcher_job(job, func, samples, *args):
+    """
+    Spawns a tree of jobs to avoid overloading the number of jobs spawned by a single parent.
+
+    :param function func: Function to spawn dynamically, passes one sample as first argument
+    :param list samples: Array of samples to be batched
+    :param list args: any arguments to be passed to the function
+    :return:
+    """
+    num_partitions = 100
```
hannes-ucsc: Make `num_partitions` a keyword argument defaulting to 100.

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def is_exe(f):
+        return os.path.isfile(f) and os.access(f, os.X_OK)
+
+    fpath, fname = os.path.split(program)
+    if fpath:
+        if is_exe(program):
+            return program
+    else:
+        for path in os.environ["PATH"].split(os.pathsep):
```
hannes-ucsc: Prefer `'` over `"`

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,54 @@
+import os
+import subprocess
+
+
+def which(program):
+    """
+    Determines if a program exists
```
hannes-ucsc: Add `FIXME: replace with bd2k.util.processes.which`

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,54 @@
...
+
+    return None
+
+
+def docker_call(tool, parameters=None, work_dir='.', env=None, sudo=False, outfile=None):
+    """
+    Calls Docker, passing along parameters and tool.
+
+    :param str tool: Name of the Docker image to be used (e.g. quay.io/ucsc_cgl/samtools)
+    :param list parameters: Parameters to be passed to the tool
```
hannes-ucsc: `lis[str]`

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    base_docker_call = ['docker', 'run',
+                        '--log-driver=none',
+                        '--rm',
+                        '-v', '{}:/data'.format(work_dir)]
+    if sudo:
+        base_docker_call.insert(0, 'sudo')
+    if env:
+        for e in env:
+            base_docker_call += ['-e', '{}={}'.format(e, env[e])]
```
hannes-ucsc: use `base_docker_call .extend()`

use `env.iteritems()`

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,54 @@
...
+    :param str work_dir: Directory to mount into the container via `-v`
+    :param dict env: Environemnt variables to be added (e.g. dict(JAVA_OPTS='-Xmx15G'))
+    :param bool sudo: If True, prepends `sudo` to the docker call
+    :param file outfile: Pipe output of Docker call to file handle
+    :return int: Exit Code
+    """
+    base_docker_call = ['docker', 'run',
+                        '--log-driver=none',
+                        '--rm',
+                        '-v', '{}:/data'.format(work_dir)]
```
hannes-ucsc: Shouldn't we use abspath(work_dir)

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,54 @@
...
+    if sudo:
+        base_docker_call.insert(0, 'sudo')
+    if env:
+        for e in env:
+            base_docker_call += ['-e', '{}={}'.format(e, env[e])]
+    if outfile:
```
hannes-ucsc: redundant

src/toil_scripts/lib/test/test_files.py
```diff
@@ -0,0 +1,59 @@
...
+
+def test_mkdir_p(tmpdir):
+    import os
+    from toil_scripts.lib.files import mkdir_p
+    work_dir = str(tmpdir)
+    dir_path = os.path.join(work_dir, 'test')
+    assert os.path.exists(work_dir)
+    mkdir_p(work_dir)
+    mkdir_p(dir_path)
+    assert os.path.isdir(dir_path)
```
hannes-ucsc: Clean up!

hannes-ucsc:
Many of the tests don't have clean up. I'll continue looking at the test code, once that is fixed.

jvivian:
> Many of the tests don't have clean up. I'll continue looking at the test code, once that is fixed.

Py.test tmpdirs clean up after themselves. 
https://bitbucket.org/pytest-dev/pytest/issues/543/should-tmpdir-clean-up-after-itself

jvivian:
> I wonder if the generality of this method would be increased if you removed the call to wGF().

I thought about that, but the intended use case involves tarballs from disparate jobs (otherwise you'd just use `tarball_files`).  Since it's a job function, I either need it to return the FileStoreId of the consolidated tarball, or it needs to accept a `func` argument to call a child job function, but I prefer the former.

jvivian:
@hannes-ucsc — Pinging merge master.

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,51 @@
+import os
+import subprocess
+
+
+def which(program):
```
hannes-ucsc: Add FIXME: replace with bd2k.util.processes.which

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,79 @@
...
+    :param str prefix: Optional prefix for files in tarball
+    """
+    if fpaths is None:
+        fpaths = []
+    with tarfile.open(os.path.join(work_dir, tar_name), 'w:gz') as f_out:
+        for fpath in fpaths:
+            f_out.add(fpath, arcname=prefix + os.path.basename(fpath))
+
+
+def move_to_output_dir(output_dir, filepaths=None):
```
hannes-ucsc: rename to move_files and swap arguments

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+    Moves files from the working directory to the output directory.
+
+    :param str output_dir: Output directory
+    :param list[str] filepaths: Filepaths to move
+    """
+    if filepaths is None:
```
hannes-ucsc: Who would call this method with None? Empty list yes, but None? If None is an actual use case, then make the for loop conditional rather than assigning empty list simply to skip the for loop.

src/toil_scripts/lib/test/test_programs.py
```diff
@@ -0,0 +1,9 @@
+def test_which():
+    from toil_scripts.lib.programs import which
+    assert which('python')
+
+
+def test_docker_call(tmpdir):
+    from toil_scripts.lib.programs import docker_call
+    work_dir = str(tmpdir)
+    docker_call(work_dir=work_dir, parameters=['--help'], tool='quay.io/ucsc_cgl/samtools')
```
hannes-ucsc: Test the efficacy of the env and outfile arguments.

src/toil_scripts/lib/test/test_urls.py
```diff
@@ -0,0 +1,46 @@
...
+    with open(fpath, 'wb') as fout:
+        fout.write(os.urandom(1024))
+    # Upload file
+    s3_dir = 's3://cgl-driver-projects/test'
+    s3am_upload(fpath=fpath, s3_dir=s3_dir, s3_encryption_key_path=key_path)
+    os.remove(fpath)
+    # Download the file
+    url = 'https://s3-us-west-2.amazonaws.com/cgl-driver-projects/test/output_file'
+    download_url(url=url, work_dir=work_dir, s3_encryption_key_path=key_path)
+    assert os.path.exists(os.path.join(work_dir, 'output_file'))
```
hannes-ucsc: Assert that output file and input file are identical.

src/toil_scripts/lib/urls.py
```diff
@@ -0,0 +1,147 @@
...
+    new_key = hashlib.sha256(master_key + url).digest()
+    assert len(new_key) == 32, 'New key is invalid and is not 32 characters: {}'.format(new_key)
+    return new_key
+
+
+def _s3am_with_retry(c, *args):
+    """
+    Calls S3AM upload with retries
+
+    :param int c: Number of cores to pass to upload/download slots
```
hannes-ucsc: Change c to `num_cores`

src/toil_scripts/lib/urls.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def download_url(url, work_dir='.', name=None, s3_encryption_key_path=None):
+    """
+    Downloads URL, can pass in file://, http://, s3://, or ftp://
+
+    :param str url: URL to download from
+    :param str work_dir: Directory to download file to
+    :param str name: Name of output file, if None, basename of URL is used
+    :param str s3_encryption_key_path: Path to the 32-byte encryption key if stored in S3
```
hannes-ucsc:
```
s/if stored in S3/if `url` points at file in S3 and that file uses SSE-C
```

hannes-ucsc:
Some of my comments from this morning weren't addressed. It's a waste of my time for me to go back and check all the outdated line comments and verify if they were indeed addressed. From now on, we will use a different policy: Don't squash your commits until the review passes. When the review passes, squash and force push. Then I will merge.

jvivian:
> Who would call this method with None? Empty list yes, but None? If None is an actual use case, then make the for loop conditional rather than assigning empty list simply to skip the for loop.

That approach was the highest rated answer to avoid supplying an empty list: http://stackoverflow.com/questions/366422/what-is-the-pythonic-way-to-avoid-default-parameters-that-are-empty-lists

In this case, I should just remove the default argument, as it doesn't make sense to supply it. 

> Some of my comments from this morning weren't addressed. [...]  From now on, we will use a different policy: [...]

Fine with me.

hannes-ucsc:
> highest rated answer 

The SO answer makes sense if you really need a list. Here you don't. You can just shortcut the loop. This is a no brainer. It's less code and it is more efficient.

hannes-ucsc:
> Who would call this method with None?

That was not a rhetorical question. Is calling this method without the 2nd argument or with a 2nd argument set to None. Really a use case?

src/toil_scripts/lib/files.py
```diff
@@ -21,7 +21,7 @@ def mkdir_p(path):
             raise
 
 
-def tarball_files(tar_name, work_dir='.', prefix='', fpaths=None):
+def tarball_files(tar_name, work_dir='.', prefix='', seperator='', fpaths=None):
```
hannes-ucsc: Why? And spelling.

jvivian:
> That was not a rhetorical question. Is calling this method without the 2nd argument or with a 2nd argument set to None. Really a use case?

Nope, which is why I removed it in the latest commit. 
https://github.com/BD2KGenomics/toil-scripts/pull/190/commits/973d2603e031dfbc607208cf0bdf55a69f92af4d#diff-2c38858d5e3b6346421b556b0f5e6a26R41

jvivian:
> Why? And spelling.

You asked for a separator argument I thought:
screen shot 2016-03-24 at 11 13 30 am

jvivian:
Nevermind, i realized you meant that the separator would be part of the prefix argument, my mistake.

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,78 @@
...
+    :return str: FileStoreID
+    """
+    work_dir = job.fileStore.getLocalTempDir()
+    # Retrieve output file paths to consolidate
+    tar_paths = []
+    for fname, file_store_id in fname_to_id.iteritems():
+        p = job.fileStore.readGlobalFile(file_store_id, os.path.join(work_dir, fname + '.tar.gz'))
+        tar_paths.append((p, fname))
+    # I/O
+    output_name = 'foo.tar.gz'
```
fnothaft: Why is this hardcoded?

src/toil_scripts/lib/jobs.py
fnothaft: Unless this number is a fixed constraint in toil, I would expose this as a parameter. At the very least, I'd add a comment to explain why 100 was chosen.

jvivian:
> Why is this hardcoded?

It's a `job` function, so the name is irrelevant as it just returns a FileStoreId.

jvivian:
> Unless this number is a fixed constraint in toil, I would expose this as a parameter. At the very least, I'd add a comment to explain why 100 was chosen.

@hannes-ucsc and I talked about this yesterday — Adding it as a parameter gets in the way of passing `*args` and it should be transparent to the user.  I agree about your second point though, i'll add an explanatory comment.

fnothaft:
@jvivian Can you respond on the inline diffs instead of the main PR comment thread? It makes the disparate threads of conversation a bit easier to follow...

jvivian:
> Can you respond on the inline diffs instead of the main PR comment thread? It makes the disparate threads of conversation a bit easier to follow...

Oops, my apologies.

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,78 @@
...
```
fnothaft: > It's a job function, so the name is irrelevant as it just returns a FileStoreId.

Ah, fair. Maybe drop a comment here to explain then. It makes sense upon explanation, but it wasn't obvious to me on the first pass of reading it, so I imagine that someone else will read over this code later and be confused.

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,53 @@
...
+    :param list[str] parameters: Parameters to be passed to the tool
+    :param str work_dir: Directory to mount into the container via `-v`
+    :param dict[str,str] env: Environemnt variables to be added (e.g. dict(JAVA_OPTS='-Xmx15G'))
+    :param bool sudo: If True, prepends `sudo` to the docker call
+    :param file outfile: Pipe output of Docker call to file handle
+    :return int: Exit Code
+    """
+    base_docker_call = ['docker', 'run',
+                        '--log-driver=none',
+                        '--rm',
```
fnothaft: Can we either omit `--rm` or make it optional? It conflicts with some of the options we use for running containers in the Spark/ADAM pipeline. `--net=host`, I think...?

src/toil_scripts/lib/urls.py
```diff
@@ -0,0 +1,147 @@
...
+    :return str: Path to the downloaded file
+    """
+    file_path = os.path.join(work_dir, name) if name else os.path.join(work_dir, os.path.basename(url))
+    if s3_encryption_key_path:
+        _download_encrypted_file(url, file_path, s3_encryption_key_path)
+    elif url.startswith('s3:'):
+        _download_s3_url(file_path, url)
+    else:
+        subprocess.check_call(['curl', '-fs', '--retry', '5', '--create-dir', url, '-o', file_path])
+    assert os.path.exists(file_path)
```
fnothaft: OOC, why would we expect this assert to fire? I assume that all of the constituent methods will throw if downloading fails?

src/toil_scripts/lib/files.py
jvivian: Sure thing

src/toil_scripts/lib/programs.py
jvivian: I've added it as an argument

src/toil_scripts/lib/urls.py
jvivian: gp, i'll remove it.

src/toil_scripts/lib/urls.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+    Uploads a file to s3 via S3AM
+    For SSE-C encryption: provide a path to a 32-byte file
+
+    :param str fpath: Path to file to upload
+    :param str s3_dir: Ouptut S3 path. Format: s3://bucket/[directory]
+    :param int num_cores: Number of cores to use for up/download with S3AM
+    :param str s3_encryption_key_path: (OPTIONAL) Path to 32-byte key to be used for SSE-C encryption
+    """
+    assert s3_dir.startswith('s3://'), 'Format of s3_dir (s3://) is incorrect: {}'.format(s3_dir)
```
fnothaft: As we were discussing earlier on the 1.0 PR, this might make more sense as a `ValueError` instead of an assert.

src/toil_scripts/lib/urls.py
```diff
@@ -0,0 +1,147 @@
...
+    s3_dir = os.path.join(s3_dir, os.path.basename(fpath))
+    if s3_encryption_key_path:
+        base_url = 'https://s3-us-west-2.amazonaws.com/'
```
fnothaft: Make region optional/open a ticket to revisit later? I had to go and run in `us-east-1` a while back, and it was a hassle to go through and modify code away from our `us-west-2` bias... ;)

src/toil_scripts/lib/urls.py
```diff
@@ -0,0 +1,147 @@
...
+
+def download_url_job(job, url, name=None, s3_encryption_key_path=None):
+    """Job version of `download_url`"""
+    work_dir = job.fileStore.getLocalTempDir()
+    fpath = download_url(url, work_dir=work_dir, name=name, s3_encryption_key_path=s3_encryption_key_path)
+    return job.fileStore.writeGlobalFile(fpath)
+
+
+def s3am_upload(fpath, s3_dir, num_cores=1, s3_encryption_key_path=None):
+    """
```
fnothaft: Might provide a reciprocal documentation link here to `s3am_upload_job`.

src/toil_scripts/lib/urls.py
```diff
@@ -0,0 +1,147 @@
...
+
+    :param str file_path: Path to file
+    :param str url: S3 URL
+    """
+    from boto.s3.connection import S3Connection
+    s3 = S3Connection()
+    try:
+        parsed_url = urlparse(url)
+        if not parsed_url.netloc or not parsed_url.path.startswith('/'):
+            raise RuntimeError("An S3 URL must be of the form s3:/BUCKET/ or "
```
fnothaft: s/RuntimeError/ValueError/g

src/toil_scripts/lib/urls.py
```diff
@@ -0,0 +1,147 @@
...
+
+    :param str url: URL to be downloaded
+    :param str file_path: Output path to file
+    :param str key_path: Path to the 32-byte key file
+    """
+    # Grab master key
+    with open(key_path, 'r') as f:
+        key = f.read()
+    if len(key) != 32:
+        raise RuntimeError('Invalid Key! Must be 32 bytes: {}'.format(key))
```
fnothaft: Since we're printing this to the log, I wanted to make sure this is a public key. Sorry if it's a silly question—I'm not familiar with the S3 encryption business.

fnothaft: I'd also probably make this a `ValueError`.

jvivian: Interesting point — the assertion is only raised though if the key _isn't_ 32 bytes, which would by definition make it an invalid key.

fnothaft: Yeah, I could imagine something though where you wrote 31 bytes of the key into the file, and you're just missing the last byte of the key, or you wrote the key and padded with an extra 0 byte. That'd still be an inadvisable security compromise, IMO.

That being said, I don't feel strongly about this, but given my druthers, I would write out something like `"Bad key in {key_file_path}. Must be 32 bytes!"`.

src/toil_scripts/lib/urls.py
jvivian: Is the proper syntax for this? 

```
if not s3_dir.startswith('s3://'):
    raise ValueError('message')
```

 One nice thing about assertion is they're concise one-liners.

src/toil_scripts/lib/urls.py
jvivian: Damn western-centric developers!  I think if we plan on adding it, I may as well just do it now.

jvivian: Sidenote: We had discussed plans of changing the encryption method from the `https://` URL to the s3 URL `s3://bucket/key`, but we've scrapped the reencryption plan as @briandoconnor plans to move us to a different storage system.

src/toil_scripts/lib/urls.py
jvivian: Can you clarify?  Not sure what you mean by documentation link.

src/toil_scripts/lib/urls.py
jvivian: Changed.

src/toil_scripts/lib/urls.py
jvivian: I agree, changed.

jvivian: I've replicated this behavior in the `_generate_unique_key` function.

src/toil_scripts/lib/urls.py
fnothaft: Correct.

> One nice thing about assertions is they're concise one-liners.

Yeah, scala has a nice `require(...)` built-in that is the equivalent of `assert` for Java's `IllegalArgumentException`.

src/toil_scripts/lib/urls.py
hannes-ucsc: The latest unstable release of s3am now has --sse-key-is-master which causes it to derive the per-file key for you. It does not hard-code the region but determines it automatically from the bucket. I would prefer to use that here and eliminate the derivation code, at least for this function.

src/toil_scripts/lib/urls.py
hannes-ucsc: -1 this is an important post-condition that should be asserted.

src/toil_scripts/lib/jobs.py
hannes-ucsc: We tried but abandoned it since keyword arguments to job functions are awkward in Toil. But yes, do explain why 100 was chosen.

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,78 @@
...
+    try:
+        os.makedirs(path)
+    except OSError as exc:
+        if exc.errno == errno.EEXIST and os.path.isdir(path):
+            pass
+        else:
+            raise
+
+
+def tarball_files(tar_name, work_dir='.', prefix='', fpaths=None):
```
hannes-ucsc: Is fpaths=None an actual use case? Who would want to create an empty archive?

src/toil_scripts/lib/urls.py
fnothaft: Sorry for the confusion. I wasn't suggesting that it should be removed (for the exact reason you stated), but rather that we document why this might happen (if there is an explicit reason that we expect this assert might fire).

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,53 @@
...
+                        '-v', '{}:/data'.format(os.path.abspath(work_dir))]
+    if sudo:
+        base_docker_call.insert(0, 'sudo')
+    if env:
+        for e, v in env.iteritems():
+            base_docker_call.extend(['-e', '{}={}'.format(e, v)])
+    docker_command = base_docker_call + [tool] + parameters
+    subprocess.check_call(docker_command, stdout=outfile)
+    return docker_command
```
hannes-ucsc: Bad. See my comment on the corresponding test.

fnothaft: Just tacking on to this to add that I would log the command that is run.

src/toil_scripts/lib/test/test_programs.py
```diff
@@ -0,0 +1,22 @@
...
+    tool = 'quay.io/ucsc_cgl/samtools'
+    docker_call(work_dir=work_dir, parameters=parameter, tool=tool)
+    # Test outfile
+    fpath = os.path.join(work_dir, 'test')
+    with open(fpath, 'w') as fout:
+        docker_call(work_dir=work_dir, parameters=parameter, tool=tool, outfile=fout)
+    assert os.path.getsize(fpath) > 0
+    # Test env
+    cmd = docker_call(work_dir=work_dir, parameters=parameter, tool=tool, env=dict(JAVA_OPTS='-Xmx15G'))
+    assert '-e JAVA_OPTS=-Xmx15G' in ' '.join(cmd)
```
hannes-ucsc:
```
docker_call( tool='ubuntu', env=dict(foo=bar), parameters='echo -n $foo'.split(), outfile='foobar')
assertEquals( open('foobar').read(), 'bar' )
```

Kills the `outfile=` and `env=` birds with a single, more rigorous stone, and without the awkwardness of returning something from `docker_call()`.

hannes-ucsc:
Besides the line notes, why aren't there any changes to the user scripts? Refactoring the existing scripts  to actually **use** the library is part of this PR.

src/toil_scripts/lib/test/test_programs.py
hannes-ucsc: `parameters='sh -c echo $foo'.split()`

fnothaft:
> Besides the line notes, why aren't there any changes to the user scripts? Refactoring the existing scripts to actually use the library is part of this PR.

+1, this should be in before merge.
However, I think the script changes will be pretty enormous, so I think it is good to make sure that we're all in agreement on the library functions _before_ we\* refactor the scripts.

\* Royal we. Although, let me know if you'd like any help refactoring the adam/gatk pipeline @jvivian.

src/toil_scripts/lib/urls.py
jvivian: I've added it back in, but don't know the specific condition why check_call would return successfully but there wouldn't be a file there, so I'm not sure exactly how to document it.

src/toil_scripts/lib/urls.py
jvivian: > The latest unstable release of s3am now has --sse-key-is-master which causes it to derive the per-file key for you. 

Awesome! Thanks for incorporating that. 

> I would prefer to use that here and eliminate the derivation code, at least for this function.

Should I list the S3AM version / commit that introduced this feature in the docstring?  That version will have to make it's way onto people's toil-box images before they can use this function.

src/toil_scripts/lib/urls.py
fnothaft:
```
# we always expect an exception to be thrown if the download fails,
# but the existence of the output file is an important invariant, so we check anyways
```

src/toil_scripts/lib/files.py
jvivian: No — This is identical to a previous issue you brought up.  I've removed the default `=None` and moved the argument:  

`def tarball_files(tar_name, fpaths, work_dir='.', prefix=''):`

src/toil_scripts/lib/urls.py
fnothaft: > Should I list the S3AM version / commit that introduced this feature in the docstring? That version will have to make it's way onto people's toil-box images before they can use this function.

+1, I would list the first release where it appeared.

src/toil_scripts/lib/test/test_programs.py
jvivian: Good idea, thanks.

src/toil_scripts/lib/programs.py
jvivian: I've corrected this given Hannes's improvement to the corresponding test and I've logged the docker_command.

src/toil_scripts/lib/test/test_programs.py
jvivian: While that command works from my terminal, I cannot get "$foo" to be expressed when run from subprocess — I've tried putting single quotes around it, running it with shell=True, and using curly braces.  If i run it with shell=True, stdout returns an empty string.  If i run it with shell=False, it returns $foo instead of bar.   

The only way i could get it to work was by directly passing the env variable through subprocess (not docker), and executing with shell=True. 

```
new_env = os.environ.copy()
new_env['foo'] = bar
p = subprocess.Popen("docker run --rm --log-driver=none ubuntu echo -n $foo", shell=True, stdout=subprocess.PIPE, env=new_env)
p.communicate()
>>> ('bar', None)
```

Any suggestions?

jvivian:
> Besides the line notes, why aren't there any changes to the user scripts? Refactoring the existing scripts to actually use the library is part of this PR.

-1

Respectfully, that doesn't make any sense to me. As @fnothaft said, the user scripts haven't been refactored to use the common libraries because they haven't stabilized.  We can allow that to happen in this PR, _then_ go through the process of refactoring the user scripts, but to me, it makes more sense to merge this PR and then assign individual tickets to refactor the pipelines to not only use the common library, but  clean up the code and improve things in general.  

This approach is also more flexible: Audrey/Frank/Michael can refactor the ADAM pipelines and I can refactor the other ones. Review comments pertaining to a pipeline's refactoring will also be contained in a single PR. 

> +1, this should be in before merge.
However, I think the script changes will be pretty enormous [...]

I thought we agreed on favoring smaller PRs?  https://github.com/BD2KGenomics/toil-scripts/issues/111#issuecomment-195092784 

I don't see any gain by tacking on user script refactoring into this PR, although I'm happy to hear arguments about why that'd be the case.

src/toil_scripts/lib/test/test_programs.py
jvivian: @arkal to the rescue — suggested `printenv foo` which I got to work in the unittest.

jvivian:
Tests fail due to outdated version of S3AM on jenkins slave — tests pass on my own machine after `pip install --pre -U s3am`

hannes-ucsc:
Adding a bunch of functions without changing the code to actually call them is not a refactoring.  How do you know that the added common function is actually usable in the individual pipelines? Let's be real, speculatively adding some stuff without validation is a pointless exercise.

How about we split this PR up, one PR per common function? I want to see the common function and the call site changes in a single PR. That is very important to me because it's the only way we can be reasonably sure that the function signature works for all or at least a majority of the pipelines.

We could keep the build infrastructure changes (jenkins.sh) here but exclude the actual functions before merging. Or we could merge the infrastructure changes with the first per-function PR.

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,59 @@
...
+    :param bool sudo: If True, prepends `sudo` to the docker call
+    :param file outfile: Pipe output of Docker call to file handle
+    :return int: Exit Code
+    """
+    base_docker_call = ['docker', 'run',
+                        '--log-driver=none',
+                        '--rm',
+                        '-v', '{}:/data'.format(os.path.abspath(work_dir))]
+    if rm:
+        base_docker_call.extend(['--rm'])
```
hannes-ucsc: append() for individual elements.

src/toil_scripts/lib/programs.py
hannes-ucsc: ... and you need to remove it here.

src/toil_scripts/lib/test/test_programs.py
```diff
@@ -0,0 +1,19 @@
...
+    from toil_scripts.lib.programs import docker_call
+    work_dir = str(tmpdir)
+    parameter = ['--help']
+    tool = 'quay.io/ucsc_cgl/samtools'
+    docker_call(work_dir=work_dir, parameters=parameter, tool=tool)
+    # Test outfile
+    fpath = os.path.join(work_dir, 'test')
+    with open(fpath, 'w') as f:
+        docker_call(tool='ubuntu', env=dict(foo='bar'), parameters=['printenv', 'foo'], outfile=f)
+    assert open(fpath).read().strip() == 'bar'
```
hannes-ucsc: Removing strip() and comparing against 'bar\n' would be a stronger assertion. It's also shorter.

src/toil_scripts/lib/urls.py
hannes-ucsc: 1.1 is the release where it appeared.

https://github.com/BD2KGenomics/cgcloud/commit/f851502a386f2df001f4863767f900b6255a3b5f

src/toil_scripts/lib/programs.py
jvivian: Addressed.

src/toil_scripts/lib/test/test_programs.py
jvivian: Addressed.

jvivian:
If I have to choose between splitting up the PR by function versus keeping it all in this PR I'd choose the latter.  

Once @hannes-ucsc and @fnothaft sign off on the library functions, I'll begin refactoring.

fnothaft:
> If I have to choose between splitting up the PR by function versus keeping it all in this PR I'd choose the latter.

I am a topic branch kind of person, so I would agree to keeping it all in a single PR.

fnothaft:
> Once @hannes-ucsc and @fnothaft sign off on the library functions, I'll begin refactoring.

Also, this has my sign off.

hannes-ucsc:
I convinced @jvivian that it is better to split this up. One PR per function or group of related functions. But each PR refactors all pipelines, except the ADAM stuff. Would that scenario have your support, @fnothaft?

fnothaft:
> I convinced @jvivian that it is better to split this up. One PR per function or group of related functions. But each PR refactors all pipelines, except the ADAM stuff. Would that scenario have your support, @fnothaft?

SGTM!

jvivian:
@fnothaft To keep you on your toes, we're changing our approach yet again.  Once I have CI setup for the RNA-seq pipeline ( #214 ), I will refactor the rnaseq_cgl pipeline to incorporate all the library functions in _this_ PR.  

That way pipeline refactoring can be done on individual tickets, but the library functions will have been tested. Sorry for the indecisiveness.

fnothaft:
> That way pipeline refactoring can be done on individual tickets, but the library functions will have been tested. Sorry for the indecisiveness.

That SGTM. I am somewhat hesitant to refactor the ADAM pipelines before we complete our runs (or at least, before we have the mocks done—see #206), and I haven't done any refactoring yet, so indecisiveness is a-OK with me.

jvivian:
This has been simplified two two commits.  One for the common library functions and one for the refactoring of the cgl pipeline to use said library functions.    Jenkins CI tests both the library functions as well as the pipeline. 

@fnothaft @almussel — If you guys have changes or suggestions to any of the library functions let me know now, I remember Hannes mentioning some additions to Docker call that you guys were using and if we can add it now it'd be preferable.

jvivian:

jvivian:
@hannes-ucsc

src/toil_scripts/lib/__init__.py
```diff
@@ -0,0 +1,67 @@
...
+            result.extend(flatten(el))
+        else:
+            result.append(el)
+    return result
+
+
+# FIXME: replace with bd2k.util humanization function
+def sizeof_fmt(num, suffix='B'):
+    """
+    Humanize input
```
hannes-ucsc: needs (doc)tests

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,77 @@
...
+    try:
+        os.makedirs(path)
+    except OSError as exc:
+        if exc.errno == errno.EEXIST and os.path.isdir(path):
+            pass
+        else:
+            raise
+
+
+def tarball_files(tar_name, fpaths, work_dir='.', prefix=''):
```
hannes-ucsc: paths or file_paths

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,77 @@
...
+    :param str work_dir: Current working directory
+    :param str prefix: Optional prefix for files in tarball
+    """
+    with tarfile.open(os.path.join(work_dir, tar_name), 'w:gz') as f_out:
+        for fpath in fpaths:
+            arcname = prefix + os.path.basename(fpath)
+            f_out.add(fpath, arcname=arcname)
+
+
+def move_files(filepaths, output_dir):
```
hannes-ucsc: paths or file_paths

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,77 @@
...
+        else:
+            raise
+
+
+def tarball_files(tar_name, fpaths, work_dir='.', prefix=''):
+    """
+    Creates a tarball from a group of files
+
+    :param str tar_name: Name of tarball
+    :param list[str] fpaths: File paths to include in the tarball
```
hannes-ucsc: Absolute or relative or both?

hannes-ucsc: paths or file_paths

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,77 @@
...
+    :param str work_dir: Current working directory
```
hannes-ucsc: It's the location of the tarball, not the current working directory.

The interaction between the actual current working directory, work_dir and the input paths (absolute or relative) worries me slightly. At the very least it should be documented.

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,77 @@
...
+        for fpath in fpaths:
+            arcname = prefix + os.path.basename(fpath)
+            f_out.add(fpath, arcname=arcname)
+
+
+def move_files(filepaths, output_dir):
+    """
+    Moves files from the working directory to the output directory.
+
+    :param str output_dir: Output directory
```
hannes-ucsc: Similar concerns regarding absolute vs. relative paths as above. 

For example, what happens if I pass "x" for this.

The behavior should be documented, inputs verified, if necessary, and verifications tested if necessary.

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,77 @@
...
+        dest = os.path.join(output_dir, os.path.basename(fpath))
+        shutil.move(fpath, dest)
+
+
+def consolidate_tarballs_job(job, **fname_to_id):
+    """
+    Combine the contents of separate tarballs into one.
+    Subdirs within the tarball will be named the keys in **fname_to_id
+
+    :param dict[str,str] fname_to_id: Key word arguments of the form: file_name_prefix=FileStoreID
```
hannes-ucsc: I wouldn't use ** for this parameter. What if my prefix is "foo-bar". That's not a valid identifier in Python and can't be used as a keyword arg. The fallback of `consolidate_tarballs_job(…,**{'foo-bar': …})` is pretty ugly.

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,77 @@
...
+    :return str: FileStoreID
```
hannes-ucsc: Does this shorthand actually work? Does PyCharm pick it up? You can find out by hovering over a reference to that function or by writing a invocation and appending `.`. If that pops up a list of the methods in `str`, you should be good. Otherwise, use `:rtype:` separately.

hannes-ucsc: `The file store ID of the generated tarball`

src/toil_scripts/lib/files.py
```diff
@@ -0,0 +1,77 @@
...
+    out_tar = os.path.join(work_dir, output_name)
+    # Consolidate separate tarballs into one
+    with tarfile.open(os.path.join(work_dir, out_tar), 'w:gz') as f_out:
+        for tar, fname in tar_paths:
+            with tarfile.open(tar, 'r') as f_in:
+                for tarinfo in f_in:
+                    with closing(f_in.extractfile(tarinfo)) as f_in_file:
+                        tarinfo.name = os.path.join(output_name, fname, os.path.basename(tarinfo.name))
+                        f_out.addfile(tarinfo, fileobj=f_in_file)
+    return job.fileStore.writeGlobalFile(out_tar)
```
hannes-ucsc: So we never got streaming to work?

src/toil_scripts/lib/jobs.py
```diff
@@ -0,0 +1,22 @@
+from toil_scripts.lib import partitions
+
+
+def sample_batcher_job(job, func, samples, *args):
```
hannes-ucsc: Nothing in here is specific to samples. I think the name should be more general.

The signature is like that of the functional `map()`. So maybe we should call this one `map_job`? And `samples` should be  called `inputs`.

hannes-ucsc: Also mention for which inputs sizes it would be appropriate or necessary to use this function.

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,58 @@
...
+
+    return None
+
+
+def docker_call(tool, parameters=None, work_dir='.', rm=True, env=None, sudo=False, outfile=None):
+    """
+    Calls Docker, passing along parameters and tool.
+
+    :param str tool: Name of the Docker image to be used (e.g. quay.io/ucsc_cgl/samtools)
+    :param list[str] parameters: Parameters to be passed to the tool
```
hannes-ucsc: `Command line arguments to be passed to the tool`

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,58 @@
...
+    :param str work_dir: Directory to mount into the container via `-v`
```
hannes-ucsc: Mention /data

src/toil_scripts/lib/programs.py
```diff
@@ -0,0 +1,58 @@
...
+    :param bool rm: Set to True to pass `--rm` flag.
+    :param dict[str,str] env: Environment variables to be added (e.g. dict(JAVA_OPTS='-Xmx15G'))
+    :param bool sudo: If True, prepends `sudo` to the docker call
+    :param file outfile: Pipe output of Docker call to file handle
+    :return int: Exit Code
```
hannes-ucsc: This returns None.

src/toil_scripts/lib/urls.py
hannes-ucsc: Why is this still a RuntimeError?

src/toil_scripts/lib/files.py
jvivian: Agreed, I've edited both the docstring and argument for clarity.

src/toil_scripts/lib/files.py
jvivian: Can I check absolute paths via a simple assert or do I need to raise a custom IllegalArgumentException?

src/toil_scripts/lib/files.py
jvivian: This is streaming — the contents of the tarfile are never unpacked to the directory, but loaded directly into the new tarball.

src/toil_scripts/lib/files.py
jvivian: Yup that doesn't work, i'll add add an `:rtype:` for all functions.

src/toil_scripts/lib/files.py
jvivian: Should I just use a dictionary then?

src/toil_scripts/lib/files.py
hannes-ucsc: True. I meant streaming the tar file to the job store using writeGlobalFileStream(). Tar is a streamable file format, AFAIK, for both creation of an archive and extraction from an archive.

jvivian:
@hannes-ucsc

PR #199, opened 2016-03-18
Move from %s to %r for formatting command strings.

fnothaft:


src/toil_scripts/adam_uberscript/adam_uberscript.py
```diff
@@ -496,10 +496,10 @@ def collect_realtime_metrics(params, threshold=0.5):
 
                                 try:
                                     check_call(cmd)
-                                    log.info("Successfully terminated instance via %s.",
+                                    log.info("Successfully terminated instance via %r.",
```
hannes-ucsc:
```
>>> 'The foo is %s' % ' '.join(['a bar','with', 'spaces'])
'The foo is a bar with spaces' # bad
>>> 'The foo is %r' % ' '.join(['a bar','with', 'spaces'])
"The foo is 'a bar with spaces'" # just as bad
>>> 'The foo is %r' % ['a bar','with', 'spaces']
"The foo is ['a bar', 'with', 'spaces']" # good
>>> 'The foo is %s' % ['a bar','with', 'spaces']
"The foo is ['a bar', 'with', 'spaces']" # that's because repr([]) == str([]) but that might not be true for other types so %r is preferred
```

hannes-ucsc:
Now I have all the time on the world to be pedantic ...

hannes-ucsc:
So basically, if you do %r with a list you shouldn't use join() on the list.

PR #205, opened 2016-03-22
Cleanup HDFS on Spark failure (resolves #105, fixes #194)

### LINKED ISSUE #105: Spark jobs should clean up HDFS on failure
We should wrap all Spark jobs that write to HDFS so that if they fail (catch exception from `call_docker`), we clean up the output they wrote and rethrow the exception. If not, we can get job failure loops that we can never exit.

### LINKED ISSUE #194: File names for intermediate files on standalone Spark cluster conflict between samples
All output should go into an HDFS subdirectory. This will help with #105 because it would let us do a recursive rm on the directory for emergency clean-up. Unused intermediate files should still be removed eagerly on a file-by-file basis.

almussel:
Resolves #105. Also includes string processing cleanup in adam pipeline

### DIFF
--- a/src/toil_scripts/adam_pipeline/adam_preprocessing.py
+++ b/src/toil_scripts/adam_pipeline/adam_preprocessing.py
@@ -145,38 +145,26 @@ def remove_file(masterIP, filename, sparkOnToil):
 
 # FIXME: unused parameter sparkOnToil
 
-def download_data(masterIP, inputs, sparkOnToil):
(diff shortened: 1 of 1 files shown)


hannes-ucsc:
Doesn't this also address #194?

fnothaft:
LGTM. I think this does resolve #194 as well. Thanks @almussel!

hannes-ucsc:
Ignore the Jenkins failures for now. Until #190 is merged builds will always fail.

hannes-ucsc:
+1 from me

PR #261, opened 2016-05-02
Make toil-scripts pip installable (resolves #260)

### LINKED ISSUE #260: Make toil-scripts pip installable

jvivian:
Create setup.py
Add setup.cfg that points to README
Set version in version.py so it can be reached by Makefile and setup.py
Update README for installation and entrypoints
Create Makefile to handle PyPi distribution, tests, and cleaning
Modify Jenkins.sh to use Makefile

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
-/.project
 *.pyc
-*~
-*#*
 *.log
-
+/src/*.egg-info
+.cache/
+.eggs/
+test-report.xml
+__pycache__
+venv/
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include version.py
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,103 @@
+# Copyright (C) 2015 UCSC Computational Genomics Lab
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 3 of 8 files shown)


hannes-ucsc:
When this PR goes into the end-game, it should probably be squashed.

jvivian:
```
Submitting dist/toil-scripts-0.1.tar.gz to https://pypi.python.org/pypi
Upload failed (403): You are not allowed to edit 'toil-scripts' package information
error: Upload failed (403): You are not allowed to edit 'toil-scripts' package information
make: *** [pypi] Error 1
Build step 'Execute shell' marked build as failure
Recording test results
```
@hannes-ucsc 
I think we need to setup the toil-scripts repo in PyPi before this build will succeed.

hannes-ucsc:
> I think we need to setup the toil-scripts repo in PyPi before this build will succeed.

Just for posterity: I remember this now. We don't include 'register' in the setup.py command invocation in Makefile because we want the package to be owned by me and Jenkins only be a maintainer. If Jenkins ran `register`, Jenkins would own the package. The fix is to run `register` manually once, and add `jenkins_ucsc` as a maintainer to the resulting package so it can `upload` to it.

hannes-ucsc:
There is your first release:

https://pypi.python.org/pypi?:action=display&name=toil-scripts&version=2.0a1.dev145

I smoke-tested via 

```
rm -rf foo ; virtualenv foo && foo/bin/pip install toil-scripts
foo/bin/toil-rnaseq
```

and it works. If you want me to merge this, just remove `needs_work` and ping me.

jvivian:
@hannes-ucsc — Squashed and rebased. All tests pass and PyPi push (pip installed in a virtualenv).

PR #295, opened 2016-06-07
Fix root ownership of docker output files (resolves #273)

### LINKED ISSUE #273: Modify Docker Run to fix root issue

jvivian:
@hannes-ucsc — I think we discussed a long time ago removing the exception statements in docker_call, but they've somehow reappeared.  Would you like me to remove them in this PR?

### DIFF
--- a/src/toil_scripts/lib/__init__.py
+++ b/src/toil_scripts/lib/__init__.py
@@ -1,4 +1,3 @@
-# FIXME: replace with bd2k.util.iterables.flatten
 def flatten(x):
     """
     Flattens a nested array into a single list
@@ -12,31 +11,6 @@ def flatten(x):
         else:
             result.append(el)
     return result
-
-
-# FIXME: replace with bd2k.util humanization function
-def sizeof_fmt(num, suffix='B'):
-    """
-    Humanize input
-    >>> sizeof_fmt(1024)
-    '1.0KB'
-    >>> sizeof_fmt(1024**2)
-    '1.0MB'
-    >>> sizeof_fmt(1024**3)
-    '1.0GB'
-    >>> sizeof_fmt(1024**4)
-    '1.0TB'
-
-    :param int num: number to be converted to human-readable string
-    :param str suffix: Size suffix; defaults to B for byte
-    :return: Human-readable string
-    :rtype: str
-    """
-    for unit in ['', 'K', 'M', 'G', 'T', 'P', 'E', 'Z']:
-        if abs(num) < 1024.0:
-            return "%3.1f%s%s" % (num, unit, suffix)
-        num /= 1024.0
-    return "%.1f%s%s" % (num, 'Y', suffix)
 
 
 def partitions(l, partition_size):
--- a/src/toil_scripts/lib/programs.py
+++ b/src/toil_scripts/lib/programs.py
@@ -1,34 +1,9 @@
 import os
 import subprocess
 import logging
+from bd2k.util.exceptions import panic
 
 _log = logging.getLogger(__name__)
-
-# FIXME: replace with bd2k.util.processes.which
-def which(program):
-    """
-    Determines if a program exists
-
-    :param str program: Name of program to check
-    :return: Path to program or None
-    :rtype: str
-    """
-
-    def is_exe(f):
-        return os.path.isfile(f) and os.access(f, os.X_OK)
-
-    fpath, fname = os.path.split(program)
-    if fpath:
-        if is_exe(program):
-            return program
-    else:
-        for path in os.environ['PATH'].split(os.pathsep):
-            path = path.strip('"')
-            exe_file = os.path.join(path, program)
-            if is_exe(exe_file):
-                return exe_file
-
-    return None
 
 
 def mock_mode():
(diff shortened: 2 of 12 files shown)


hannes-ucsc:
Catching CalledProcessError is harmful and needs to be removed. Catching OSError may improve usability, but only if errno is checked for ENOENT or whatever errno is set to in case `docker` can't be found on PATH. Without that extra check, catching OSError is also harmful. So either remove both except: clauses or remove the CalledProcessError one and amend the OSError one with the extra check.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
Jenkins, test this please.

hannes-ucsc:
This is really useful since it allows for removing wrapper.sh from most of cgl-docker-lib's containers, right?

This goes back to the idea of images not adding cruft around the tool.

jvivian:
> This is really useful since it allows for removing wrapper.sh from most of cgl-docker-lib's containers, right? This goes back to the idea of images not adding cruft around the tool.

Correct, this is a great change because containers no longer have to come from cgl-docker-lib to be compatible with Toil (which can't properly perform cleanup if the container creates directories owned by root).  `ENTRYPOINT`'s for tools can now just call the binary directly.  

The wrapper is nice for using the tool *outside* of Toil, but that's not our primary motivation for tool containerization.

jvivian:

src/toil_scripts/lib/programs.py
```diff
@@ -123,11 +123,12 @@ def docker_call(tool,
                 return subprocess.check_output(docker_call)
             else:
                 subprocess.check_call(docker_call)
-
-    except subprocess.CalledProcessError:
-        raise RuntimeError('docker command returned a non-zero exit status. Check error logs.')
-    except OSError:
-        raise RuntimeError('docker not found on system. Install on all nodes.')
+    finally:
```
hannes-ucsc: Sorry for not catching this earlier: The problem with `finally:` is that an exception raised there will mask any exception that occurred in `try:`

```
>>> try:
...     raise ValueError()
... finally:
...     raise RuntimeError()
...     
  File "<input>", line 4, in <module>
RuntimeError
```

There are various ways around this, none of them pretty, caused by the fact that https://docs.python.org/2/library/sys.html#sys.exc_info() is backed by a single thread-local variable.

```
>>> from bd2k.util.exceptions import panic
>>> try:
...     raise ValueError()
... except:
...     with panic():
...         raise RuntimeError()
... else:
...     pass
...     
  File "<input>", line 5, in <module>
  File "<input>", line 2, in <module>
ValueError
```

The easiest would be to omit the `try: finally:` altogether and accept the fact that on exceptions, permissions might be wrong. But I have the feeling that that won't work for us. bd2k-python-lib's panic() context manager is the next best thing I can think of. It would require that you duplicate the permission fix in both the `except:` (enclosed in `panic()`) and the `else:` (without `panic()`). That can be alleviated by extracting that code into a helper function.

jvivian:
Thanks for the info, that's an interesting problem I didn't know about. 

If I wanted to use `panic` would I need to add bd2k-python-lib as an `install_requires`?  I'm guessing the answer is yes, but it seems brittle if the bd2k-python-lib version and the Toil version mismatch, since we allow the Toil version installed to be a range #299

hannes-ucsc:
And by the way. There are a bunch of FIXME's in toil-scripts that are about replacing duplicated functionality with an import from bd2k-python-lib. IOW, `panic()` wouldn't be the only thing that we pull in from bd2k-python-lib. 

Getting tired of typing bd2k-python-lib, let's abbreviate it as BPL.

src/toil_scripts/lib/programs.py
```diff
@@ -1,35 +1,10 @@
 import os
 import subprocess
 import logging
+from bd2k.util.exceptions import panic
 
 _log = logging.getLogger(__name__)
 
-# FIXME: replace with bd2k.util.processes.which
```
hannes-ucsc: BPL's which is a generator, so the old invocations  `which('…')` have to be replaced with `next(which('…'),None)`.

src/toil_scripts/lib/__init__.py
```diff
@@ -1,4 +1,3 @@
-# FIXME: replace with bd2k.util.iterables.flatten
 def flatten(x):
```
hannes-ucsc: BPL's flatten is more general, you should be able to do this:

```
from bd2k.util.iterables import flatten as bpl_flatten

def flatten(x):
    return list(bpl_flatten(x))
```

It will be slightly more efficient than the current implementation since only one list is being created.

You can leave the test case in-place.

hannes-ucsc: Sorry that should be 

```
return list(bpl_flatten((x,))
```

or actually, there is `concat()`, which I love.

```
from bd2k.util.iterables import concat

def flatten(x):
    return list(concat(x))
```

jvivian: Hm, I can't get your version of `flatten` to perform what the current version does:

Common lib version:
```
In [43]: flatten( ('cat',('foo', 'bar'),[[['tacobell']]] ))
Out[43]: ['cat', 'foo', 'bar', 'tacobell']
``` 

BPL version:
```
from bd2k.util.iterables import concat

def flatten(x):
    return list(concat(x))
In [48]: flatten( ('cat',('foo', 'bar'),[[['tacobell']]] ))
Out[48]: ['cat', ('foo', 'bar'), [[['tacobell']]]]
```
Other version you posted:
```
from bd2k.util.iterables import flatten as bpl_flatten

def flatten(x):
    return list(bpl_flatten((x,))
In [51]: flatten( ('cat',('foo', 'bar'),[[['tacobell']]] ))
Out[51]: ['cat', ('foo', 'bar'), [[['tacobell']]]]
```

src/toil_scripts/lib/programs.py
jvivian: good catch, thanks

src/toil_scripts/lib/__init__.py
hannes-ucsc: Sorry for wasting your time. I forgot that neither concat() nor flatten() are recursive.

jvivian:
I anticipate needing an additional commit since I bumped Toil version to 3.2.0a2 wrt to front-end fixes like excising `os.remove()`.  Besides that and `--restart` -> `--resume` I'm not sure what else will affect the pipelines.

jvivian:
> Maybe we should not bump Toil version in this PR.

I'm a bit torn on this — master is "cutting edge" so there's no issue with production, at the same time it's in some sense developing against a moving target. I'm fine sticking with 3.1.6 if you're cool with that. 

> Also, c47a818 looks really suspicious since rebasing and conflict resolution do not introduce new commits.

When I made those adjustments during merge resolution, it failed. I'll clean it up.

hannes-ucsc:
We'll bump Toil to the latest 3.2.0a1 pre-release in a follow-on PR right after this PR is merged. Just to avoid dragging it out even longer. I feel that there might be additional repercussions of upgrading to Toil unstable. And we can deal with those in a separate PR. 

Ideally your branch should start at the tip of master and end in the commit whose hash is now b47c3b1, but will become a different hash due to rebasing.

jvivian:

PR #299, opened 2016-06-07
Remove Toil from install_requires (resolves #292)

### LINKED ISSUE #292: setup.py should not declare Toil as install_requires
So I've thought about this intensely. I think it is wrong for toil-script's `setup.py` to refer to toil via `install_requires`. There is always the danger that when the user installs toil-scripts (especially when installing into a virtualenv), that a different version of Toil is pulled in via that dependency. It may be a different version of Toil than what the user has on the system, and it will be installed without extras while the user's version was installed with extras.

Instead `setup.py` should contain an explicit version check, which I will outline shortly.

jvivian:
Thanks for doing the legwork on this Hannes.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,8 +19,22 @@ import sys
 from setuptools import find_packages, setup
 from setuptools.command.test import test as TestCommand
 from version import version
+from pkg_resources import parse_version
 
-toil_version = '3.1.4'
+# Toil version check -- Raise warning instead of using intall_requires to avoid virtualenv conflicts
+toil_min_version = '3.1'
+toil_max_version = '3.2'
+
+try:
+    from toil.version import version as toil_version
+except ImportError:
+    raise RuntimeError('Cannot find Toil version {}.x for Toil. Read about installing Toil at: '
+                       'http://toil.readthedocs.io/en/latest/installation.html'.format(toil_min_version))
+
+if not parse_version(str(toil_min_version)) <= parse_version(toil_version) < parse_version(toil_max_version):
+    raise RuntimeError('Need version {}.x of Toil. Read about installing Toil at: '
+                       'http://toil.readthedocs.io/en/latest/installation.html'
+                       ''.format(toil_min_version))
 
 kwargs = dict(
     name='toil-scripts',
@@ -30,7 +44,6 @@ kwargs = dict(
     author_email='cgl-toil@googlegroups.com',
     url="https://github.com/BD2KGenomics/toil-scripts",
     install_requires=[
-        'toil==' + toil_version,
         'boto==2.38.0', # FIXME: Make an extra
         'tqdm==3.8.0', # FIXME: Remove once ADAM stops using it (superfluous import)
         'pyyaml==3.11'],
--- a/setup.py
+++ b/setup.py
@@ -22,8 +22,8 @@ from version import version
 from pkg_resources import parse_version
 
 # Toil version check -- Raise warning instead of using intall_requires to avoid virtualenv conflicts
-toil_min_version = '3.1'
-toil_max_version = '3.2'
+toil_min_version = '3.1.6'
+toil_max_version = '3.2.0'
 
 try:
     from toil.version import version as toil_version
@@ -32,9 +32,9 @@ except ImportError:
                        'http://toil.readthedocs.io/en/latest/installation.html'.format(toil_min_version))
 
 if not parse_version(str(toil_min_version)) <= parse_version(toil_version) < parse_version(toil_max_version):
-    raise RuntimeError('Need version {}.x of Toil. Read about installing Toil at: '
+    raise RuntimeError('Need Toil version within range [{},{}). Read about installing Toil at: '
                        'http://toil.readthedocs.io/en/latest/installation.html'
-                       ''.format(toil_min_version))
+                       ''.format(toil_min_version, toil_max_version))
 
 kwargs = dict(
     name='toil-scripts',
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,6 @@ red=\033[0;31m
 develop: check_venv
 	$(pip) install -e .$(extras)
 clean_develop: check_venv
-	- $(pip) uninstall -y toil
 	- rm -rf src/*.egg-info
 
 
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -11,6 +11,7 @@ export PATH=$PATH:${PWD}/bin
 # Create Toil venv
 virtualenv venv
 . venv/bin/activate
+pip install toil==3.1.4
 make develop
 make test
 make clean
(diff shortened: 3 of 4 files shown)


setup.py
```diff
@@ -19,8 +19,20 @@
...
 from version import version
+from pkg_resources import parse_version
+
+# Toil version check -- Raise warning instead of using intall_requires to avoid virtualenv conflicts
+toil_minor_version = '3.1'
+next_minor_version = toil_minor_version[:2] + str(int(toil_minor_version[3]) + 1)
+try:
+    from toil.version import version as toil_version
+except ImportError:
+    raise RuntimeError('Cannot find Toil. Read about installing Toil at "toil.readthedocs.org"')
```
hannes-ucsc: More specific link, remove quotes and make it a proper URL with http:// etc.

setup.py
```diff
@@ -19,8 +19,20 @@
...
+
+if not parse_version(toil_minor_version) <= parse_version(toil_version) < parse_version(next_minor_version):
+    raise RuntimeError('Need version {}.x of Toil. '
+                       'Read about installing Toil at: "toil.readthedocs.org"'.format(toil_minor_version))
```
hannes-ucsc: Same as above.

setup.py
```diff
@@ -19,8 +19,20 @@
 from setuptools import find_packages, setup
 from setuptools.command.test import test as TestCommand
 from version import version
+from pkg_resources import parse_version
+
+# Toil version check -- Raise warning instead of using intall_requires to avoid virtualenv conflicts
+toil_minor_version = '3.1'
+next_minor_version = toil_minor_version[:2] + str(int(toil_minor_version[3]) + 1)
```
hannes-ucsc: Use parse_version. It exists so you don't have to hand-code brittle version parsing like that.

setup.py
hannes-ucsc: And by more specific I mean pointing at the installation section.

setup.py
jvivian: Totally agree — I tried to do that first but `parse_version` just returns a tuple with a string for the 2nd value which seems just as messy to update by 1.  I was hoping it'd return a useful object like `urlparse` does.

hannes-ucsc: If you use ipython in a venv, you get the old non-setuptools version of parse_version, the one that returns a dumb tuple. Either don't use ipython for your interactive tests, or use the technique explained [here](https://coderwall.com/p/xdox9a/running-ipython-cleanly-inside-a-virtualenv) to launch ipython inside a venv:

```
(bd2k:/__me__/)hannes-mbpr:~ hannes $ virtualenv venv
New python executable in venv/bin/python
Installing setuptools, pip, wheel...done.
(bd2k:/__me__/)hannes-mbpr:~ hannes $ . venv/bin/activate
(venv)(bd2k:/__me__/)hannes-mbpr:~ hannes $ pip install ipython
… snip …
(venv)(bd2k:/__me__/)hannes-mbpr:~ hannes $ python -c 'import IPython; IPython.terminal.ipapp.launch_new_instance()'
Python 2.7.10 (default, Oct 23 2015, 19:19:21) 
In [1]: from pkg_resources import parse_version

In [2]: parse_version('1.0')
Out[2]: <Version('1.0')>
```

Note how it's not a tuple anymore.

setup.py
```diff
@@ -19,8 +19,22 @@
 from setuptools import find_packages, setup
 from setuptools.command.test import test as TestCommand
 from version import version
+from pkg_resources import parse_version
 
-toil_version = '3.1.4'
+# Toil version check -- Raise warning instead of using intall_requires to avoid virtualenv conflicts
+toil_minor_version = 3.1
```
hannes-ucsc: Using floats for discrete math is suicide. It may work in this very particular case, since format() is truncating the tiny imprecisions but it is a really bad programming practice that I don't want to see anywhere in our code.

http://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency/3730040#3730040

If you use ipython in a venv, you get the old non-setuptools version of parse_version, the one that returns a dumb tuple. Either don't use ipython for your interactive tests, or use the technique explained [here](https://coderwall.com/p/xdox9a/running-ipython-cleanly-inside-a-virtualenv) to launch ipython inside a venv:

```
(bd2k:/__me__/)hannes-mbpr:~ hannes $ virtualenv venv
New python executable in venv/bin/python
Installing setuptools, pip, wheel...done.
(bd2k:/__me__/)hannes-mbpr:~ hannes $ . venv/bin/activate
(venv)(bd2k:/__me__/)hannes-mbpr:~ hannes $ pip install ipython
… snip …
(venv)(bd2k:/__me__/)hannes-mbpr:~ hannes $ python -c 'import IPython; IPython.terminal.ipapp.launch_new_instance()'
Python 2.7.10 (default, Oct 23 2015, 19:19:21) 
In [1]: from pkg_resources import parse_version

In [2]: parse_version('1.0')
Out[2]: <Version('1.0')>
```

Note how it's not a tuple anymore.

hannes-ucsc: The augmented parse_version wouldn't simplify incrementing the minor version either. Just omit the arithmetic and hard-code both version limits in the code. That's still better than a brittle or overly complicated attempt at solving it programmatically.

```
toil_min_version='3.1.5'
toil_max_version='3.2'
```

jvivian: Wow, that's good to know. When can we switch over to [Unums](http://www.johngustafson.net/presentations/Unums2.0.pdf)? 

This new `parse_version` object has a method that returns the version, base_version, which just returns a string of '3.1', which sort of puts me back at square one.

setup.py
```diff
@@ -19,8 +19,22 @@
 from setuptools import find_packages, setup
 from setuptools.command.test import test as TestCommand
 from version import version
+from pkg_resources import parse_version
 
-toil_version = '3.1.4'
+# Toil version check -- Raise warning instead of using intall_requires to avoid virtualenv conflicts
+toil_min_version = '3.1'
```
hannes-ucsc: I think you want to mention a concrete version here, 3.1.6 to be specific. There is no reason to allow, say 3.1.4 because 3.1.6 is fully backwards-compatible with 3.1.4 and the upgrade should be straightforward.

That reminds me, I should bump Toil on toil-jenkins-slave in cgcloud.

setup.py
```diff
@@ -19,8 +19,22 @@
...
+toil_max_version = '3.2'
```
hannes-ucsc: How can we convey that the interval is half-open, i.e. that max version is exclusive?

setup.py
hannes-ucsc: Sorry, I meant toil-box.

setup.py
heuermh: For better or worse, this is the spec that maven (and other JVM build tools, by extension) use
https://maven.apache.org/enforcer/enforcer-rules/versionRanges.html

jvivian: I'll amend the message to provide a concrete interval range.

jvivian:

Issue #310, opened 2016-06-20
Remove author tags from scripts


fnothaft:
As per discussion with @hannes-ucsc @benedictpaten on scrum today.

PR #311, opened 2016-06-20
Clean up source file headers (resolves #309, #310)

### LINKED ISSUE #309: Attach Apache 2 license header to all source files
See http://www.apache.org/licenses/LICENSE-2.0.html#apply

fnothaft:
Makes two changes:

* Attaches the ASF suggested license file header to all source files. This license header attributes copyright to the Regents of the University of California. Let me know if someone would like the copyright changed; I have this scripted, so it is a quick change. We'd probably want to add a `NOTICES.txt` file if making the copyright more specific.
* Removes `@author` tags, as per discussion with @benedictpaten @hannes-ucsc on today's scrum.


heuermh:
+1

setup.py
```diff
@@ -1,4 +1,5 @@
-# Copyright (C) 2015 UCSC Computational Genomics Lab
+#
+# Copyright 2015–2016, Regents of the University of California
```
hannes-ucsc: Should we do this in Toil, too?

hannes-ucsc:
Build fails with

```
pip2.7 install -e .
Obtaining file:///mnt/ephemeral/workspace/toil-scripts
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/mnt/ephemeral/workspace/toil-scripts/setup.py", line 23, in <module>
        from version import version
      File "version.py", line 2
    SyntaxError: Non-ASCII character '\xe2' in file version.py on line 2, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

setup.py
fnothaft: I generally prefer the `Regents of the University of California` copyright—esp. if you have contributors from multiple UCs—but I'm not picky.

fnothaft:
> Build fails with <snip>

Hah! I got cocky and used an en dash instead of a hyphen in 2015–2016. Will fix.

fnothaft:
Rebased and addressed buld issue.

PR #316, opened 2016-06-22
Remove hardcoded test path (resolves #315)

### LINKED ISSUE #315: Factor out hardcoded `/mnt/ephemeral` references
This came to mind while reviewing [a line](https://github.com/BD2KGenomics/toil-scripts/blob/23bc4702c5b2007ccdfcbf29ee5c132723728a70/src/toil_scripts/rnaseq_cgl/test/test_rnaseq_cgl.py#L13) that was referenced in #314. We've got a variety of places in the repo where we've hardcoded `/mnt/ephemeral` into our scripts, under the assumption that we'll always be running on a cgcloud configured AWS box. This is fine and dandy, until you're running outside of AWS.

I feel like we could roll this into some sort of `get_ephemeral_directory` function that would:

* Look for the existence of a TOIL_EPHEMERAL environment variable
* Barring that, look for the existence of /mnt/ephemeral
* And finally, barring that, call to https://docs.python.org/2/library/tempfile.html or something

jvivian:


src/toil_scripts/lib/__init__.py
```diff
@@ -31,6 +35,16 @@ def partitions(l, partition_size):
         yield l[i:i + partition_size]
 
 
+def get_work_directory(env_variable='TOIL_WORKDIR', backup_dir='/mnt/ephemeral'):
```
hannes-ucsc: This method should create a uniquely named temporary subdirectory in the first two cases.

s/backup/fallback otherwise it sounds like a directory for making backups.

I need to be convinced me of the value of being able to override `env_variable` and `backup_dir`/`fallback_dir`. We may want to tweak the implementation in the future and a complicated signature would make these tweaks harder.

jvivian:

jvivian:
@hannes-ucsc -- Ready for review

src/toil_scripts/lib/__init__.py
```diff
@@ -31,6 +36,17 @@ def partitions(l, partition_size):
...
+    try:
+        return os.path.join(os.environ['TOIL_WORKDIR'], str(uuid4()))
+    except KeyError:
+        fallback_dir = '/mnt/ephemeral'
+        if os.path.exists(fallback_dir):
+            return os.path.join(fallback_dir, str(uuid4()))
+        else:
+            return tempfile.mkdtemp()
+
+
```
hannes-ucsc:
```
def get_work_directory():
    try:
        basedir = os.environ['TOIL_WORKDIR']
    except KeyError:
        basedir = '/mnt/ephemeral'
        if not os.path.exists(basedir):
            basedir = None
    return tempfile.mkdtemp(dir=basedir)
```

is more concise, safer (since the code in `try:` can't raise a false positive) and more uniform wrt the naming of the output directory.

src/toil_scripts/rnaseq_cgl/test/test_rnaseq_cgl.py
```diff
@@ -2,31 +2,31 @@
...
+from toil_scripts.lib import get_work_directory
+
 
 def test_rnaseq_cgl(tmpdir):
-    subdir = '/mnt/ephemeral/toil-scripts/rnaseq/'
-    work_dir = os.path.join(subdir, str(uuid4()))
-    mkdir_p(work_dir)
-    create_config_and_manifest(work_dir)
+    workdir = get_work_directory()
+    mkdir_p(workdir)
```
hannes-ucsc: Not necessary anymore.

src/toil_scripts/lib/__init__.py
hannes-ucsc: But need to add docstring specifying the post-condition, i.e. directory exists when function returns.

jvivian: Oh nice, I should've looked at `mkdtemp`'s arguments.   

One thing: `mkdtemp` uses `os.mkdir`, which doesn't create directories recursively, so should I move the `os.path.exists()` out a level in case the `os.environ['TOIL_WORKDIR']` doesn't exist?

hannes-ucsc: When `TOIL_WORKDIR` is defined, it should point to an existing directory and if it doesn't I would want to know rather than mask that problem with `mkdir_p`. I think the same applies to the hard-coded fallback.

jvivian:
@hannes-ucsc - Ready for review

src/toil_scripts/lib/__init__.py
```diff
@@ -31,6 +35,22 @@ def partitions(l, partition_size):
         yield l[i:i + partition_size]
 
 
+def get_work_directory():
+    """
+    Creates work directory that exists when function returns
```
hannes-ucsc:
```
Creates a a working directory at a location that is suitable for storing files 
temporarily. The caller is responsible for removing the directory.

:return: The path to the directory
:rtype: str
```

jvivian:

hannes-ucsc:
+1. Squash, then merge.

PR #324, opened 2016-06-27
Refactor bwa (resolves #320, resolves #297, resolves #322)

### LINKED ISSUE #297: Refactor BWA pipeline to use config/manifest

### LINKED ISSUE #320: Require only reference for BWA
Annoying and user unfriendly to require all the BWA indices and fai as standard input. Will now be generated if not present.

### LINKED ISSUE #322: Add single-end support for BWA
Pipeline currently only supports alignment of fastq pairs, but bwa-kit can align single-ended data.

jvivian:


fnothaft:
I will give this a review...

jvivian:
@fnothaft I can ping you once I remove the needs work label and tests pass.

fnothaft:
> @fnothaft I can ping you once I remove the needs work label and tests pass.

That would be superb!

jvivian:
@fnothaft — I spread out the history to make it more readable, but its not perfect, so if something looks odd check the [finished product](https://github.com/BD2KGenomics/toil-scripts/blob/b812b86d067e470bdd0adcc2d9d44c7323ad4576/src/toil_scripts/batch_alignment/bwa_alignment.py).

fnothaft:
Thanks @jvivian; I'll make a pass shortly.

jvivian:
Thanks! — I'm happy to squash commits or make changes to make the review easier.

fnothaft:
@jvivian that would be great actually. I think it might make sense to squash down:

* 86d36b3 and 7c5eb6f into a single commit that resolves #297 
* 1ce621f, 8613ea3, and 60348dd into a single commit that resolves #322
* cab3f8b, 040cc38, and 	edc4ea3 into a single commit that resolves #320 
* I'd probably also squash all of the others down into a single "cleanup" commit.

jvivian:
Thanks @fnothaft I'll ping you when I've squashed and rebased.

jvivian:
@fnothaft —Almost as asked. I had a demon commit (53f) that gave me merge conflicts with missing code anytime I moved commits around it or tried to squash it.

fnothaft:
Demon commit; nice! I'll take a look.

src/toil_scripts/adam_gatk_pipeline/align_and_call.py
```diff
@@ -178,13 +178,13 @@ def static_dag(job, uuid, rg_line, inputs):
...
-    bwa = job.wrapJobFn(download_shared_files,
+    inputs.rg_line = rg_line
+    inputs.output_dir = 's3://{s3_bucket}/alignment{dir_suffix}'.format(**args)
+    bwa = job.wrapJobFn(download_reference_files,
                         inputs,
                         [uuid,
                          's3://{s3_bucket}/{sequence_dir}/{uuid}_1.fastq.gz'.format(**args),
-                         's3://{s3_bucket}/{sequence_dir}/{uuid}_2.fastq.gz'.format(**args)],
-                        's3://{s3_bucket}/alignment{dir_suffix}'.format(**args),
-                        rg_line).encapsulate()
```
fnothaft: What happens to `rg_line` now?

fnothaft: Er. Nevermind. I see that `rg_line` is now being passed through `inputs`.

fnothaft: Why isn't `rg_line` just in `inputs` to start with? If we're refactoring this, mightn't we factor out passing `rg_line` as a parameter?

src/toil_scripts/batch_alignment/README.md
```diff
@@ -1,47 +1,146 @@
 ## University of California, Santa Cruz Genomics Institute
-### GATK-compatible Alignment
+### Guide: Running the BWA Pipeline using Toil
+
+This guide attempts to walk the user through running this pipeline from start to finish. If there are any questions
+please contact John Vivian (jtvivian@gmail.com). If you find any errors or corrections please feel free to make a
```
fnothaft: Ixnay on user IDs in docs, nyet? See #310.

src/toil_scripts/batch_alignment/README.md
```diff
@@ -1,47 +1,146 @@
...
+If Toil is already installed globally (true for CGCloud users), or there are global dependencies (like Mesos),
+use virtualenv's `--system-site-packages` flag.
+ 
 ## Dependencies
 
 This pipeline has been tested on Ubuntu 14.04, but should also run on other unix based systems.  `apt-get` and `pip`
-often require `sudo` privilege, so if the below commands fail, try prepending `sudo`.  If you do not have sudo 
-privileges you will need to build these tools from source, or bug a sysadmin (they don't mind). 
+often require `sudo` privilege, so if the below commands fail, try prepending `sudo`.  If you do not have `sudo` 
+privileges you will need to build these tools from source, or bug a sysadmin about how to get them (they don't mind).
```
fnothaft: At the risk of being joyless, I would remove:

> (they don't mind)

src/toil_scripts/batch_alignment/README.md
```diff
@@ -1,47 +1,146 @@
...
 
-This pipeline produces a BAMFILE for a given sample.
+## Example Config
 
-## Running / Help
 
-It is recommended to use the associated launch scripts which provide default arguments needed to run the pipeline. 
-It is likely that the job store positional argument, `--workDir`, and `--output-dir` arguments will need to be modified.
-To run a pipeline after dependencies have been installed, simply:
+    # BWA Alignment Pipeline configuration file
```
fnothaft: Put in verbatim block? \`\`\`

src/toil_scripts/batch_alignment/README.md
```diff
@@ -1,47 +1,146 @@
...
+    # Required: The library entry to go in the BAM read group.
+    library: Illumina
+    
+    # Required: Platform to put in the read group
+    platform: Illumina
+    
+    # Required: Program Unit for BAM header. Required for use with GATK.
+    program_unit: 12345
+    
+    # Required: Approximate input file size. Should be given as [integer][TGMK]
```
fnothaft: The prefixes will _likely_ be obvious to most readers, but you might want to give an example. Also, clarify whether this is base-10 (Giga) or base-2 (Gibi), bits, bytes, etc.

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,254 @@
...
 
-def download_shared_files(job, inputs, sample, output_dir, rg_line=None):
+def download_reference_files(job, inputs, samples):
     """
     Downloads shared files that are used by all samples for alignment
 
     :param JobFunctionWrappingJob job: passed automatically by Toil
-    :param JobFunctionWrappingJob job: Passed by Toil automatically
     :param Namespace inputs: Input arguments (see main)
+    :param list[list[str, list]] samples: Sample UUIDs and URLs
```
fnothaft: If this is a doubly nested list containing string/list tuples, I think it needs more of a description.
Also, I thought we preferred `:param` + `:type`?

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,254 @@
...
-    job.addChildJobFn(map_job, download_sample, samples, inputs, shared_ids, cores=1, disk=inputs.file_size)
-
-
-def download_sample(job, sample, inputs, ids):
+    shared_ids = {}
+    urls = [('amb', inputs.amb), ('ann', inputs.ann), ('bwt', inputs.bwt),
+            ('pac', inputs.pac), ('sa', inputs.sa)]
+    # If all inputs provided, download from S3
+    download_ref = job.wrapJobFn(download_url_job, inputs.ref, disk='3G')
+    faidx = job.wrapJobFn(run_samtools_faidx, download_ref.rv())
```
fnothaft: Move into conditional?

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,254 @@
...
+    if all(urls):
```
fnothaft: There's a lot going on in this set of conditionals. Could you write some line comments to steer the reader through it?

fnothaft: Ping me to rereview this stretch once there's docs.

src/toil_scripts/batch_alignment/bwa_alignment.py
hannes-ucsc: In Toil we use `:param $type $name: $description`. I love that shortcut. Can't wait to upgrade to Python 3 where you can annotate the parameters directly.

fnothaft: Ah, I had it flipped. Disregard. Thanks for the clarification, @hannes-ucsc.

> If this is a doubly nested list containing string/list tuples, I think it needs more of a description.

Still holds, though.

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,254 @@
...
-    job.fileStore.logToMaster('Parsed configuration file.')
-    job.addChildJobFn(map_job, download_sample, samples, inputs, shared_ids, cores=1, disk=inputs.file_size)
-
-
-def download_sample(job, sample, inputs, ids):
+    shared_ids = {}
+    urls = [('amb', inputs.amb), ('ann', inputs.ann), ('bwt', inputs.bwt),
+            ('pac', inputs.pac), ('sa', inputs.sa)]
+    # If all inputs provided, download from S3
+    download_ref = job.wrapJobFn(download_url_job, inputs.ref, disk='3G')
```
fnothaft: This reminds me of the famous paper "Hardcoding resource requirements considered harmful."

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,254 @@
...
-              output_file)
-
-    # Either write file to local output directory or upload to S3 cloud storage
-    job.fileStore.logToMaster('Aligned sample: {}'.format(inputs.uuid))
-
-    if inputs.output_dir.startswith('s3://'):
-        s3am_upload(output_file, inputs.output_dir, s3_key_path=inputs.ssec)
+        ids['r2'] = None
+    # Define and wire job functions
+    bam_id = job.wrapJobFn(run_bwakit, inputs.cores, ids['r1'], ids['r2'], ids['ref'], ids['fai'], ids['amb'],
```
fnothaft: There's a variety of stuff going on here that I'm just not feeling. I'm holding a squirmy baby right now so I can't type well, but I will come back with more detail.

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,254 @@
...
+    Parse manifest file
+
+    :param str manifest_path: Path to manifest file
+    :return: samples
+    :rtype: list[str, list]
+    """
+    samples = []
+    with open(manifest_path, 'r') as f:
+        for line in f.readlines():
+            if not line.isspace() and not line.startswith('#'):
```
fnothaft: Use parsing library?

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,254 @@
...
+                                               'Expected UUID\tURL1\t[URL2] (tab separated), got: {}'.format(sample))
+                uuid = sample[0]
+                urls = sample[1:]
+                for url in urls:
+                    require(urlparse(url).scheme and urlparse(url), 'Invalid URL passed for {}'.format(url))
+                samples.append([uuid, urls])
+    return samples
+
+
+def required_length(nmin, nmax):
```
fnothaft: Factor into `toil_scripts.lib`?

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -226,40 +257,43 @@ def main():
...
                             '\nDefault value: "%(default)s".')
-    group.add_argument('--sample', default=None, nargs='3', type=str,
-                       help='Space delimited sample UUID and paired fastq files in the format: uuid url1 url2')
-    parser_run.add_argument('--output-dir', default=None, help='Full path to directory or filename where '
-                                                               'final results will be output')    
-    parser_run.add_argument('-s', '--suffix', default='.bqsr', help='Additional suffix to add to the names of the output files.')
+    group.add_argument('--sample', nargs='+', action=required_length(2, 3),
+                       help='Space delimited sample UUID and fastq files in the format: uuid url1 [url2].')
+    # Print docstring help if no arguments provided
+    print sys.argv
```
fnothaft: I would remove this.

If you were going to keep it, I think @hannes-ucsc would want `repr(sys.argv)`. I would also want this to be `log.info`/`log.debug`ed, not `print`ed.

src/toil_scripts/lib/files.py
```diff
@@ -35,6 +35,13 @@ def move_files(file_paths, output_dir):
         shutil.move(file_path, dest)
 
 
+def move_file_job(job, name, file_id, output_dir):
+    """Job version of `move_files` for a single file"""
```
fnothaft: Moar docs plz.

Also, what exactly is the use case for this function? I'll dig back into the diff to see where it is invoked, but as a bystander I'm not entirely understanding the "why" for this function.

src/toil_scripts/tools/aligners.py
```diff
@@ -56,3 +56,79 @@ def run_star(job, cores, r1_id, r2_id, star_index_url):
...
+    :param str ann: FileStoreID of ann
+    :param str bwt: FileStoreID of bwt
+    :param str pac: FileStoreID of pac
+    :param str sa: FileStoreID of sa
+    :param str uuid: UUID of sample (for readgroup)
+    :param str library: library of sample (for readgroup)
+    :param str platform: platform of sample (for readgroup)
+    :param str program_unit: program unit for sample (for readgroup)
+    :param str alt: FileStoreID for alt sequence
+    :param str rg_line: rg line. Use the format "@RG\\tID:foo\\tLB:bar" ...
```
fnothaft: That format line isn't really descriptive. I'd go more for something like 

> BAM read group header line (`@RG`), as defined [on page 3 of the SAM spec](https://samtools.github.io/hts-specs/SAMv1.pdf). Tabs should be escaped, e.g., `@RG\\tID:foo\\tLB:bar...` for the read group "foo" from sequencing library "bar". Multiple `@RG` lines can be defined, but should be split by an escaped newline `\\n`, e.g., `@RG\\tID:foo\\t:LB:bar\\n@RG\\tID:santa\\tLB:cruz`.

Yes, I know it's verbose, but I lost several hours of my life several months ago to this `@RG` header business, and good documentation could save someone else that hassle.

src/toil_scripts/tools/aligners.py
```diff
@@ -56,3 +56,79 @@ def run_star(job, cores, r1_id, r2_id, star_index_url):
...
+    :param bool sort: If True, sorts the BAM
+    :param bool trim: If True, performs adapter trimming
+    :return: FileStoreID of BAM
+    :rtype: str
+    """
+    work_dir = job.fileStore.getLocalTempDir()
+    file_names = ['r1.fq.gz', 'ref.fa.fai', 'ref.fa', 'ref.fa.amb', 'ref.fa.ann',
+                  'ref.fa.bwt', 'ref.fa.pac', 'ref.fa.sa']
+    ids = [r1, ref, fai, amb, ann, bwt, pac, sa]
+    if r2:
```
fnothaft: Add in-line docs for the conditionals.

src/toil_scripts/tools/aligners.py
```diff
@@ -56,3 +56,79 @@ def run_star(job, cores, r1_id, r2_id, star_index_url):
...
+    if alt:
+        file_names.append('ref.fa.alt')
+        ids.append(alt)
+    for fileStoreID, name in zip(ids, file_names):
+        job.fileStore.readGlobalFile(fileStoreID, os.path.join(work_dir, name))
+    # Read groups
+    if rg_line:
+        rg = rg_line
+    else:
+        rg = "@RG\\tID:{0}".format(uuid)  # '\' character is escaped so bwakit gets individual '\' and 't' characters
```
fnothaft: This `@RG` line generation behavior needs to be hella documented.

src/toil_scripts/tools/aligners.py
```diff
@@ -56,3 +56,79 @@ def run_star(job, cores, r1_id, r2_id, star_index_url):
...
+        opt_args.append('-s')
+    if trim:
+        opt_args.append('-a')
+    # Call: bwakit
+    parameters = (['-t', str(threads),
+                   '-R', rg] +
+                  opt_args +
+                  ['-o', '/data/aligned',
+                   '/data/ref.fa',
+                   '/data/r1.fq.gz'])
```
fnothaft: Why don't you move the optional `parameters.append('/data/r2.fq.gz')` up here so that all of the parameter creation logic is in a single place?

src/toil_scripts/tools/aligners.py
```diff
@@ -56,3 +56,79 @@ def run_star(job, cores, r1_id, r2_id, star_index_url):
...
+    mock_bam = uuid + '.bam'
+    outputs = {'aligned.aln.bam': mock_bam}
+    if r2:
+        parameters.append('/data/r2.fq.gz')
+    docker_call(tool='quay.io/ucsc_cgl/bwakit:0.7.12--528bb9bf73099a31e74a7f5e6e3f2e0a41da486e',
+                parameters=parameters, inputs=file_names, outputs=outputs, work_dir=work_dir)
+
+    # BWA insists on adding an `*.aln.sam` suffix, so rename the output file
```
fnothaft: We don't need this file renaming bit anymore.

src/toil_scripts/tools/preprocessing.py
```diff
@@ -43,3 +43,42 @@ def cutadapt(job, r1_id, r2_id, fwd_3pr_adapter, rev_3pr_adapter):
         r1_cut_id = job.fileStore.writeGlobalFile(os.path.join(work_dir, 'R1_cutadapt.fastq'))
         r2_cut_id = None
     return r1_cut_id, r2_cut_id
+
+
+def run_bwa_index(job, ref_id):
```
fnothaft: Neither of these two functions are what I would call preprocessing. Maybe create an `indexing` module?

src/toil_scripts/tools/preprocessing.py
```diff
@@ -43,3 +43,42 @@ def cutadapt(job, r1_id, r2_id, fwd_3pr_adapter, rev_3pr_adapter):
...
+
+    :param JobFunctionWrappingJob job: passed automatically by Toil
+    :param str ref_id: FileStoreID for the reference genome
+    :return: FileStoreID for reference index
+    :rtype: str
+    """
+    job.fileStore.logToMaster('Created reference index')
+    work_dir = job.fileStore.getLocalTempDir()
+    job.fileStore.readGlobalFile(ref_id, os.path.join(work_dir, 'ref.fasta'))
+    command = ['faidx', 'ref.fasta']
```
fnothaft: `/data/ref.fasta`?

src/toil_scripts/batch_alignment/bwa_alignment.py
fnothaft: OK, squirmy baby drifted off to sleep. A variety of thoughts:

* `run_bwakit` now takes 19 parameters. This doesn't seem like a positive refactoring outcome to me, esp. since we're unpacking a dictionary to pass those 19 parameters. I understand the drive towards having just fileStore IDs passed around, but I think that having 9 fileStore IDs + many various parameters is too cumbersome.
* What's the rationale to decomposing `run_bwakit`, `s3am_upload_job`, `move_file_job` into child jobs? IMO, there's no real gain. They're not logically separate stages. You don't gain concurrency. I mean, maybe I'm being an idiot—and if I am, you should tell me—but isn't this always going to lead to the aligned BAM getting written to the filestore? If we're writing the file into S3 or to a local directory, this just doesn't make sense.
* Disk requirement should really be 3x `file_size` + index size, no?

hannes-ucsc: > 19 parameters

That's a lot. Especially since they are passed by position. Good luck maintaining/refacorting that while not being on Adderall. It is ok to pass positional aka mandatory arguments by name. I'm aware that some of these parameters are sample-specific, but the non-sample-specific ones should be extracted from the config.

The point about the config is that it should be passed to every job, and every job extracts the information it needs from it. No unpacking or translation should happen. The config can also be modified, when a job imports a static input file, the file's URL is read from the config, the URL is imported, the config entry is modified by replacing the URL with the job store file ID and the modified config is returned.

This should work:

```py
config.ref = addChildJob( wrapJobFn( import_file_job( config.ref ) ) ).rv()
```

fnothaft: > That's a lot. Especially since they are passed by position. Good luck maintaining/refacorting that while not being on Adderall.

Especially since most of the inputs have the same type... You would likely not even get a type error if you messed up the ordering.

jvivian:
Thanks for the feedback @fnothaft — working on this now.

src/toil_scripts/adam_gatk_pipeline/align_and_call.py
jvivian: > Why isn't rg_line just in inputs to start with? 

Looking at the code, I'm not sure if it's even defined anywhere in `align_and_call`. 

`rg_line` gets set from `uuid_items[1]`, which is taken from `uuid_rg` from `uuid_list`, but `uuid_list` is parsed from the manifest which has no entry for read group...  (confused yet?).

src/toil_scripts/batch_alignment/README.md
jvivian: Good catch — Technically this isn't in a script and I thought it'd be useful to provide a point of contact.  We could make a toil-scripts google group and provide that instead?

src/toil_scripts/batch_alignment/README.md
jvivian: I indented it because a ``` block wasn't working — or at least not formatting correctly in pycharm.  I think the line of hashes screwed something up.

screen shot 2016-06-29 at 10 51 45 am

src/toil_scripts/batch_alignment/bwa_alignment.py
jvivian: `shared_ids`, `faidx`, and `download_ref` are used in both forks of the conditional.

src/toil_scripts/batch_alignment/bwa_alignment.py
jvivian: > This reminds me of the famous paper "Hardcoding resource requirements considered harmful."

Agreed, but I think it's a bit overkill to ask the user for the reference size in the config...  I don't mind bumping this up to something like 5G or 10G to be on the safe side, but I've personally never used a reference bigger than hg19.fa / hg38.fa.

src/toil_scripts/batch_alignment/bwa_alignment.py
jvivian: That was left in by mistake, sorry.

src/toil_scripts/batch_alignment/bwa_alignment.py
jvivian: This is a really good point, and probably means changes to the `gatk_preprocessing` function in my current exome PR. 

@fnothaft — Any chance you'd be available on Slack today?  @hannes-ucsc suggested using a config, but I didn't want to force people to adopt my version of the config (which accesses attributes via the dot-method). Do you have issues with Hannes's suggestion or should I go for it?

src/toil_scripts/batch_alignment/bwa_alignment.py
jvivian: I could switch to `csv`, but since the manifest has comments it wouldn't be much cleaner:

```
with open(manifest_path, 'r') as tsv:
    for line in csv.reader(tsv, delimiter='\t'):
        if line and not line[0].startswith('#'):
....
````

I'd save one line by skipping the strip and split on tab.

File: src/toil_scripts/lib/files.py (hunk 14 above)
Comment (jvivian) [inline src/toil_scripts/lib/files.py]: `move_files` operates on files that have accessible file paths.  I needed a job version of that function so I could pipe the FileStoreID, returned as a promise, to `move_file_job`. 

I asked Hannes if the one-liner docstring was OK, given that `move_file` is defined right above it and the `job` versions of the URL functions follow the same convention.  I'll go ahead and change it since it you requested.

**Comment by jvivian** [comment]
@fnothaft — Individual line comments were addressed with separate commits so you could see exactly what I fixed. I can squash these as requested. 

All we need to do is hash out the best approach to `run_bwakit`'s arguments so I can make that change.  Thanks again for the thorough review!

File: src/toil_scripts/batch_alignment/bwa_alignment.py (hunk 10 above)
Comment (fnothaft) [inline src/toil_scripts/batch_alignment/bwa_alignment.py]: I +1 the config approach.

**Comment by jvivian** [comment]
@fnothaft — Ready for review / discussion

File: src/toil_scripts/batch_alignment/bwa_alignment.py (hunk 11 above)
Comment (fnothaft) [inline src/toil_scripts/batch_alignment/bwa_alignment.py]: Ah, maybe ignore it then.

File: src/toil_scripts/batch_alignment/README.md (hunk 2 above)
Comment (fnothaft) [inline src/toil_scripts/batch_alignment/README.md]: +1 to a toil-scripts user group. Possibly gitter rooms too?

File: src/toil_scripts/batch_alignment/README.md (hunk 4 above)
Comment (fnothaft) [inline src/toil_scripts/batch_alignment/README.md]: Ah, ok!

File: src/toil_scripts/batch_alignment/bwa_alignment.py (hunk 6 above)
Comment (fnothaft) [inline src/toil_scripts/batch_alignment/bwa_alignment.py]: Ping:

> If this is a doubly nested list containing string/list tuples, I think it needs more of a description.

File: src/toil_scripts/batch_alignment/bwa_alignment.py (hunk 8 above)
Comment (fnothaft) [inline src/toil_scripts/batch_alignment/bwa_alignment.py]: Ping RE: docs.

File: src/toil_scripts/batch_alignment/bwa_alignment.py (hunk 7 above)
Comment (fnothaft) [inline src/toil_scripts/batch_alignment/bwa_alignment.py]: OK! Missed that.

File: src/toil_scripts/batch_alignment/bwa_alignment.py (hunk 9 above)
Comment (fnothaft) [inline src/toil_scripts/batch_alignment/bwa_alignment.py]: That's reasonable, but add a comment describing the rationale for a number that is otherwise a magic number.

File: src/toil_scripts/adam_gatk_pipeline/align_and_call.py (hunk 1 above)
Comment (fnothaft) [inline src/toil_scripts/adam_gatk_pipeline/align_and_call.py]: Ace!

File: src/toil_scripts/batch_alignment/bwa_alignment.py [hunk 22]
```diff
@@ -1,223 +1,248 @@
...
-
-
-def download_sample(job, sample, inputs, ids):
+    shared_ids = {}
+    urls = [('amb', inputs.amb), ('ann', inputs.ann), ('bwt', inputs.bwt),
+            ('pac', inputs.pac), ('sa', inputs.sa)]
+    # If all inputs provided, download from S3
+    download_ref = job.wrapJobFn(download_url_job, inputs.ref, disk='3G')
+    faidx = job.wrapJobFn(run_samtools_faidx, download_ref.rv())
+    # If all the bwa index files are provided
```
fnothaft: Semantics, but I would note here that the FAI is considered separate from the BWA indices, so we'll check for that later.

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,248 @@
...
+    if all(urls):
+        if inputs.alt:
```
fnothaft: Note that the alt file is optional and "can't" be regenerated.

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,248 @@
...
+            urls.append(('alt', inputs.alt))
+        else:
+            shared_ids['alt'] = None
+        if inputs.fai:
+            urls.append(('fai', inputs.fai))
+        else:
+            # If just fai is absent, create it
```
fnothaft: We should probably log the missing FASTA index here.

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,248 @@
...
+            job.addChild(download_ref)
+            download_ref.addChild(faidx)
+            shared_ids['ref'] = download_ref.rv()
+            shared_ids['fai'] = faidx.rv()
+        # Download the reference, bwa index files, and fai / alt if provided
+        for name, url in urls:
+            shared_ids[name] = job.addChildJobFn(download_url_job, url).rv()
+    else:
+        # Create bwa index files since they were not provided
```
fnothaft: Ditto RE: logging.

Do we not check for the FAI in this branch? We always regenerate it?

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,248 @@
...
+        bwa_index = job.wrapJobFn(run_bwa_index, download_ref.rv())
+        job.addChild(download_ref)
+        download_ref.addChild(faidx)
+        download_ref.addChild(bwa_index)
+        shared_ids['ref'] = download_ref.rv()
+        shared_ids['fai'] = faidx.rv()
+        for x, name in enumerate(['amb', 'ann', 'bwt', 'pac', 'sa']):
+            shared_ids[name] = bwa_index.rv(x)
+        job.addFollowOnJobFn(map_job, download_sample_and_align, samples, inputs, shared_ids)
```
fnothaft: Shouldn't this not be in the conditional?

fnothaft: Also, can you document that `map_job` maps over the list elements in `samples` or something. That code gave me pause and I had to look across files to suss out what was going on.

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -1,223 +1,248 @@
...
-    output_file = os.path.join(work_dir, '{}.bam'.format(inputs.uuid))
-    os.rename(os.path.join(work_dir, 'aligned.aln.bam'),
-              output_file)
-
-    # Either write file to local output directory or upload to S3 cloud storage
-    job.fileStore.logToMaster('Aligned sample: {}'.format(inputs.uuid))
-
-    if inputs.output_dir.startswith('s3://'):
-        s3am_upload(output_file, inputs.output_dir, s3_key_path=inputs.ssec)
+        ids['r2'] = None
```
fnothaft: This block of code could use some whitespace to separate functionally separate lines of code.

src/toil_scripts/lib/files.py
fnothaft: Ah, got it.

src/toil_scripts/tools/aligners.py
```diff
@@ -56,3 +56,85 @@ def run_star(job, cores, r1_id, r2_id, star_index_url):
...
+    :param bool sort: If True, sorts the BAM
+    :param bool trim: If True, performs adapter trimming
+    :return: FileStoreID of BAM
+    :rtype: str
+    """
+    work_dir = job.fileStore.getLocalTempDir()
+    file_names = ['r1.fq.gz', 'ref.fa.fai', 'ref.fa', 'ref.fa.amb', 'ref.fa.ann',
+                  'ref.fa.bwt', 'ref.fa.pac', 'ref.fa.sa']
+    ids = [config.r1, config.ref, config.fai, config.amb, config.ann, config.bwt, config.pac, config.sa]
+    if getattr(config, 'r2', None):  # If a fastq pair was provided
```
fnothaft: Legibility nit: separate the conditionals and for loops with whitespace and move the comments to their own lines.

src/toil_scripts/tools/aligners.py
```diff
@@ -56,3 +56,85 @@ def run_star(job, cores, r1_id, r2_id, star_index_url):
...
+        file_names.insert(1, 'r2.fq.gz')
+        ids.insert(1, config.r2)
+    if getattr(config, 'alt', None):  # If an alt file was provided
+        file_names.append('ref.fa.alt')
+        ids.append(config.alt)
+    for fileStoreID, name in zip(ids, file_names):
+        job.fileStore.readGlobalFile(fileStoreID, os.path.join(work_dir, name))
+    # Read groups
```
fnothaft: This is not a useful comment. I'd go with something more like "Was a read group line provided? If not, generate this line by using the UUID blah blah blah"

src/toil_scripts/tools/aligners.py
fnothaft: Ping.

fnothaft:
Just made another pass @jvivian! Let me know your thoughts!

src/toil_scripts/tools/aligners.py
jvivian: I might've misunderstood this one — the `parameter.append` code is directly below this line now, but you could also mean move it so there's only one call to `if getattr(config, 'r2', None):` ?

src/toil_scripts/batch_alignment/bwa_alignment.py
jvivian: > Do we not check for the FAI in this branch? We always regenerate it?

I decided to just regenerate it because the logic was easier and the bottleneck here is definitely `bwa index` which takes 1-2 hours to run relative to `faidx`.  If that annoys you I'll change it.

src/toil_scripts/batch_alignment/bwa_alignment.py
jvivian: > Shouldn't this not be in the conditional?

Ouch >_<

> Also, can you document that map_job maps over the list elements in samples or something. That code gave me pause and I had to look across files to suss out what was going on.

Yup no problem.

src/toil_scripts/tools/aligners.py
jvivian: Compromise on moving comments to their own lines to separate conditionals and loops?

jvivian:
Great feedback @fnothaft , thanks!  I'll ping you once tests pass. 

I also pinged Hannes about the toil-scripts google group.

src/toil_scripts/batch_alignment/bwa_alignment.py
fnothaft: I don't feel strongly. If I had written this, then yes, I'd probably check and not regenerate the FAI if it is present. But, if not, I'd document why we're ignoring it.

src/toil_scripts/batch_alignment/bwa_alignment.py
fnothaft: >> Shouldn't this not be in the conditional?
>
> Ouch >_<

I suppose the follow up is, have you verified the refactored pipeline? :-p

src/toil_scripts/tools/aligners.py
fnothaft: What's the drawback from your side? I don't see much of a draw to omitting the whitespace, but I can be persuaded.

Word on the street is that whitespace doesn't cost extra, so you might as well use as much of it as you need.

src/toil_scripts/tools/aligners.py
fnothaft: I was thinking of something like how `opt_args` is used, but for the read pair FASTQ.

src/toil_scripts/batch_alignment/bwa_alignment.py
jvivian: I've cleaned this up and fixed the bad logic

src/toil_scripts/batch_alignment/README.md
```diff
@@ -2,8 +2,9 @@
 ### Guide: Running the BWA Pipeline using Toil
 
 This guide attempts to walk the user through running this pipeline from start to finish. If there are any questions
-please contact John Vivian (jtvivian@gmail.com). If you find any errors or corrections please feel free to make a 
-pull request.  Feedback of any kind is appreciated.
+please contact the Toil team at: Toil@googlegroups.com
```
fnothaft: Move to project level readme, no? We should have it there, and not duplicate it across all pipeline readmes.

src/toil_scripts/batch_alignment/bwa_alignment.py
```diff
@@ -54,6 +54,7 @@ def download_reference_files(job, inputs, samples):
         download_ref.addChild(bwa_index)
         for x, name in enumerate(['amb', 'ann', 'bwt', 'pac', 'sa']):
             shared_ids[name] = bwa_index.rv(x)
+    # Map_job distributes one sample in samples to the downlaod_sample_and_align functoin
```
fnothaft: Whitespace before comment for legibility.

src/toil_scripts/batch_alignment/README.md
jvivian: Oops, I'll do that now.

fnothaft:
Dropped two comments; otherwise branch is good to squash down and rebase from my side.

jvivian:
@hannes-ucsc — This has been reviewed, squashed and rebased, and is ready to merge.

hannes-ucsc:
Hmm, don't like how the commits are organized. I should be able to revert a commit and that should undo exactly one change. I bet you that 48ffd14 touches on all of the above commits.

In that sense, I'd rather have just one giant commit. But now that I have to share my Merge button superpowers, diluted beyond recognition to a mere plus or minus one, I don't want to hold this up. Your thoughts, @fnothaft?

jvivian:
I squashed the commits because I had to rebase and didn't want to
potentially deal with 33 commits worth of conflicts. I guess I should've
pinged you for review before that, but frank had already looked at them at
that point so I thought I was clear to, sorry about that.

fnothaft:
> I should be able to revert a commit and that should undo exactly one change.
>
> In that sense, I'd rather have just one giant commit. But now that I have to share my Merge button superpowers, diluted beyond recognition to a mere plus or minus one, I don't want to hold this up. Your thoughts, @fnothaft?

I strongly +1 the 1 commit = 1 change approach. That's how we run all of the bdgenomics repositories, and I'd love to have this be an official policy over in BD2KGenomics as well.

hannes-ucsc:
@jvivian, no need to apologize. Organizing changes into commits is an art, especially when it comes to large refactorings. Given what @fnothaft said, I think squashing all commits down to one would be best. I do like your detailed commit messages, so be sure to merge them all together so as to not lose that information.

jvivian:
@hannes-ucsc — Ready for review/merge

PR #333, opened 2016-06-30
Fix KeyError when samples are untarred (resolves #334)

### LINKED ISSUE #334: Hotfix RNA-seq pipeline for non-tar fastq samples
non-tar fastqs were never properly tested on 1.1 release and a key error arose:

http://hgwdev.cse.ucsc.edu/~jeltje/toil/R.fastq.error

jvivian:


jvivian:
@hannes-ucsc — 1.1 didn't have CI setup so there are no tests to run.  I've simply moved the key value declaration up the DAG to avoid the missing KeyError raised when samples skip the `process_sample_tar` function due to being untarred.

hannes-ucsc:
Sounds good. On branch naming: I'm still honing in on this but I think the branch name should have been: `issues/334-some-description/1.x`. If there already is a branch named `issues/334-some-description` then the above won't work and I don't have a good solution for it. 

Maybe `issues/1.x/334-some-description` but that doesn't group related branches.

jvivian:
Gotcha, I followed the schema from the previous hotfix but it'd be good to codify our implementation of gitflow and branch naming.

PR #337, opened 2016-07-01
Fix single-end labelling logic (resolves #336)

### LINKED ISSUE #336: Fix single-end flag logic in production rna-seq

jvivian:
Will bump production release to 1.3 and I'll deprecate 1.2


hannes-ucsc:
I think /issues/336-fix-single-end-logic/1.x because 1.x is the branch you are targeting.

PR #345, opened 2016-07-11
Clarify local inputs (RNA-seq) (resolves #344)

### LINKED ISSUE #344: Clarify inputs as URLs (RNA-seq)
User input requires URLs, but isn't as explicit as it should be.

- Add some logic for user input 
- Add some clarification to README
- Add some clarification in config string

jvivian:


src/toil_scripts/rnaseq_cgl/rnaseq_cgl_pipeline.py
```diff
@@ -478,6 +480,9 @@ def main():
                                                            '{}, RSEM: {}'.format(config.star_index, config.rsem_ref))
         require(config.output_dir or config.s3_output_dir, 'output-dir AND/OR s3-output-dir need to be defined, '
                                                            'otherwise sample output is not stored anywhere!')
+        for input in [x for x in [config.kallisto_index, config.star_index, config.rsem_ref] if x]:
+            require(urlparse(input).scheme, 'Input in config must have the appropriate URL prefix: '
```
hannes-ucsc: Don't you want to enforce a particular set of schemes, rather than enforce that scheme is present and non-empty.

Also, the set of supported schemes shouldn't be hard-coded in duplicate. Ideally, it should be declared in one place as a set or tuple and comments in the config template and require() calls should use that set/tuple.

jvivian: Good points, I'll do that now.

Issue #329, opened 2016-06-28
ADAM pipeline inexplicably runs twice


almussel:
When running the ADAM preprocessing pipeline on its own (as opposed to running it as part of the ADAM/GATK pipeline), the entire pipeline runs twice, despite finishing successfully the first time. This includes spinning up an entirely new Spark cluster if using Spark on Toil.

PR #328, opened 2016-06-28
Convert ADAM pipeline to Toil Spark support (resolves #325)

almussel:

Issue #326, opened 2016-06-28
Memory settings tuning for ADAM pipeline


fnothaft:
Current pipeline hits some GC issues it appears. Should adjust:

* Cache memory fraction
* Network timeout settings

Issue #325, opened 2016-06-28
Remove spark cluster support


fnothaft:
As of https://github.com/BD2KGenomics/toil/commit/6456fe81b385ac9314ec684be9fd5d41fb8d545a, the Spark support in toil-scripts is duplicative and should be removed in favor of the upstream Spark service code.

PR #351, opened 2016-07-13
Run adam-pipeline in single machine mode (resolves #210, #329, #371, #325, #326)

### LINKED ISSUE #210: Support running ADAM/GATK pipeline on local storage
* Bypass storage to S3
* Run Spark using `local`

fnothaft:
Depends on #328. I presume that we'll get #328 merged, and then sort out this one.

If anyone would like to make a review pass though, the last three commits are "the business".

This is needed for https://github.com/BD2KGenomics/cgl-docker-lib/issues/139.


src/toil_scripts/tools/spark_tools.py
```diff
@@ -128,28 +131,58 @@ def call_conductor(master_ip, src, dst, memory=None, override_parameters=None):
...
+              local_dir=None,
+              native_adam_path=None):
     """
     Invokes the ADAM container. Find ADAM at https://github.com/bigdatagenomics/adam.
 
     :param masterIP: The Spark leader IP address.
     :param arguments: Arguments to pass to ADAM.
     :param memory: Gigabytes of memory to provision for Spark driver/worker.
     :param override_parameters: Parameters passed by the user, that override our defaults.
+    :param native_adam_path: Path to ADAM executable. If not provided, Docker is used.
```
fnothaft: TODO: Document `run_local`, `local_dir`

src/toil_scripts/tools/spark_tools.py
```diff
@@ -128,28 +131,58 @@ def call_conductor(master_ip, src, dst, memory=None, override_parameters=None):
...
+    else:
+        master = ["--master",
+                  ("spark://%s:%s" % (master_ip, SPARK_MASTER_PORT)),
+                  "--conf", ("spark.hadoop.fs.default.name=hdfs://%s:%s" % (master_ip, HDFS_MASTER_PORT)),]
+        work_dir = '.'
+
+    default_params = (master + [
+            # set max result size to unlimited, see #177
+            "--conf", "spark.driver.maxResultSize=0",
+            "--conf", "spark.storage.memoryFraction=0.3",
```
fnothaft: TODO: document these parameters.

src/toil_scripts/adam_pipeline/adam_preprocessing.py
```diff
@@ -303,8 +352,8 @@ def main():
     parser_run.add_argument('--config', default='adam_preprocessing.config', type=str,
                             help='Path to the (filled in) config file, generated with "generate-config". '
                                  '\nDefault value: "%(default)s"')
-    parser_run.add_argument('--sample', help='The full s3 url of the input SAM or BAM file')
-    parser_run.add_argument('--output-dir', default=None,
+    parser_run.add_argument('--sample', help='The full s3 url/path to the input SAM or BAM file')
```
jvivian: This is a bit confusing — In the RNA-seq pipeline there's a top-level tuple of URL schemes that gets interpolated into strings that mention we use URLs as inputs.

src/toil_scripts/lib/files.py
```diff
@@ -21,18 +21,18 @@ def tarball_files(tar_name, file_paths, output_dir='.', prefix=''):
             f_out.add(file_path, arcname=arcname)
 
 
-def copy_files(file_paths, output_dir):
+def __forall_files(file_paths, output_dir, op):
```
jvivian: This is great, love this change.

src/toil_scripts/lib/files.py
```diff
@@ -49,6 +49,26 @@ def copy_file_job(job, name, file_id, output_dir):
     copy_files([fpath], output_dir)
 
 
+def move_files(file_paths, output_dir):
```
jvivian: This is a semi-dangerous function to use now that we have caching. Might want to put a warning in the docstring?

jvivian:
Made a couple nit comments, but aside from that LG2M.  You have a few magic numbers but already have line comments about adding documentation.

src/toil_scripts/lib/files.py
fnothaft: Good catch, will address.

src/toil_scripts/lib/files.py
fnothaft: Thanks! I was pretty chuffed with it myself.

src/toil_scripts/adam_pipeline/adam_preprocessing.py
fnothaft: I believe we only support S3 URLs and file paths here. Let me clean up the docs, though.

fnothaft: This is a bit different from the other paths, since we do the download using conductor.

fnothaft:
Thanks for the comments @jvivian! Just pushed fd214ba to address.

fnothaft:
@jvivian just updated the last commit to fix the test issue.

BTW, our tests rely on files that are only accessible with UCSC AWS creds (probably with --requestor-pays works too, haven't checked there). I'll open a ticket for us to resolve that.

jvivian:
All tests passed, but it looks like there's an import error in `align_and_call` that's referring to the (now deleted) uberscript dir.

fnothaft:
Fixed and pushed.

fnothaft:
@jvivian hold on merging this 30min, I just realized that I did something breathlessly stupid in my last commit.

jvivian:
@fnothaft — No sweat, just ping me.

fnothaft:
Actually fixed this time. Ping for merge after tests pass, @jvivian.

jvivian:
Thanks @fnothaft !

PR #355, opened 2016-07-14
Convert RNASeq test to TestCase and randomize output (resolves #354)

### LINKED ISSUE #354: Allow for concurrent execution of RNASeq tests
Currently the output of the RNASeq tests are hard-coded. If two or more builds are executing the same test concurrently—and this is not an unlikely scenario when building PRs on Jenkins—they will overwrite each other's output.

Te output directory in S3 should be chosen pseudo-randomly and a clean way would be to set output_dir in setupClass(). Unfortunately, the tests currently all are flat functions so we may have to convert them to methods of a unittest.TestCase subclass. The alternative is passing input_dir and output_dir around as parameters.

hannes-ucsc:
Using a class-based unit test just makes it easier to control what is initialized at which stage of the the test lifecycle. It also should open us up to provide a shared TestCase base class in toil-lib.

### DIFF
--- a/jenkins.sh
+++ b/jenkins.sh
@@ -13,9 +13,19 @@ virtualenv venv
 . venv/bin/activate
 # Adding AWS extra to get boto as required by tests
 pip install toil[aws]==3.3.0
+
+# Prepare directory for temp files
+TMPDIR=/mnt/ephemeral/tmp
+rm -rf $TMPDIR
+mkdir $TMPDIR
+export TMPDIR
+
 make develop
 make test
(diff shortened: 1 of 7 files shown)


jvivian:
LGTM — After this is merged I'll push out the release.

hannes-ucsc:
Let's leave this out of the release. I might have to tweak it a bit. My main goal is to call this from the integration tests.

hannes-ucsc:
Jenkins, test this please.

PR #392, opened 2016-07-28
Fix output_dir bug

jvivian:


jvivian:

PR #407, opened 2016-08-09
Add fastqc zip to output (resolves #404)

### LINKED ISSUE #404: add full FastQC output to toil-rnaseq
Currently, `fastqc: true` results in a single HTML file per input fastq. However, we're working on a cohort QC pipeline that uses both the fastqc_data.txt file and the separate .png image files for each sample. Can these be added to the tar file?

jvivian:


jvivian:
@fnothaft — Ready for review

fnothaft:
LGTM! Merged. Thanks @jvivian!

PR #335, opened 2016-06-30
Refactor gatk germline (resolves #280)

jpfeil:
resolves #280 #426 #425 #401 

@jvivian  @fnothaft @hannes-ucsc

PR #418, opened 2016-08-16
Finish production germline pipeline (resolves #401)

### LINKED ISSUE #401: Add clinical variant annotation to germline

jpfeil:
Adds features for the production release of the germline pipeline.
- Oncotator annotation
- Synapse download 

resolves #401 
connected to #335 

@fnothaft @jvivian


jvivian:
Thanks @jpfeil — I'll take a pass at this tonight.

src/toil_scripts/tools/variant_annotation.py
```diff
@@ -0,0 +1,85 @@
...
+    :return: Annotated VCF FileStoreID
+    :rtype: str
+    """
+    job.fileStore.logToMaster('Running Oncotator')
+    work_dir = job.fileStore.getLocalTempDir()
+    inputs = {'oncotator_db': config.oncotator_db,
+              'input.vcf': vcf_id}
+    get_files_from_filestore(job, work_dir, inputs)
+
+    # Call GATK -- HaplotypeCaller
```
fnothaft: Looks like comment is stale; I don't think we're running GATK here.

src/toil_scripts/tools/variant_annotation.py
```diff
@@ -0,0 +1,85 @@
...
+from __future__ import print_function
+import os
+
+from toil_scripts.lib.files import get_files_from_filestore
+from toil_scripts.lib.programs import docker_call
+
+
+def run_oncotator(job, vcf_id, config):
+    """
+    Runs Oncotator on VCF file. This tool only uses HG19 annotations.
```
fnothaft: Do we check here/elsewhere that the provided reference genome is an HG19 build?

src/toil_scripts/tools/variant_annotation.py
```diff
@@ -0,0 +1,85 @@
...
+    job.fileStore.logToMaster('Running GATK VariantAnnotator: {}'.format(config['uuid']))
+    work_dir = job.fileStore.getLocalTempDir()
+    references = ['genome.fa', 'genome.fa.fai', 'genome.dict']
+    inputs = {key: config[key] for key in references}
+    inputs['input.bam'] = bam_id
+    inputs['input.bam.bai'] = bai_id
+    inputs['input.vcf'] = vcf_id
+    get_files_from_filestore(job, work_dir, inputs)
+
+    # Call GATK -- HaplotypeCaller
```
fnothaft: Ditto here RE stale comment.

src/toil_scripts/gatk_germline/germline.py
```diff
@@ -1,304 +1,500 @@
...
+    :return: Updated config with shared fileStoreIDS
+    :rtype: Namespace
     """
     job.fileStore.logToMaster('Downloading shared reference files')
-    reference_names = ['genome.fa', 'phase.vcf', 'mills.vcf', 'dbsnp.vcf', 'hapmap.vcf', 'omni.vcf']
-    if config['run_bwa']:
-        bwa_references = ['amb', 'ann', 'bwt', 'pac', 'sa']
-        reference_names.extend(bwa_references)
-    for name in reference_names:
+    references = {'genome_fasta'}
```
fnothaft: Nit, but `references` seems like the wrong term here, since most of these files are indices or known variant files.

src/toil_scripts/gatk_germline/germline.py
```diff
@@ -1,304 +1,500 @@
...
+    :return: BAM and BAI FileStoreIDs
+    :rtype: tuple
+    """
+    # Determine extension on input file
+    base, ext1 = os.path.splitext(url)
+    _, ext2 = os.path.splitext(base)
+
+    # Produce or download BAM file
+    if {ext1, ext2} & {'.fq', '.fastq'}:
+        if not config.run_bwa:
```
fnothaft: Refactor these lines to use `require`.

src/toil_scripts/gatk_germline/germline.py
```diff
@@ -1,304 +1,500 @@
...
+    :param Namespace config: Configuration options for pipeline
+    :param str rg_line: RG line for BWA alignment. Default is None.
+    :return: BAM and BAI FileStoreIDs
+    :rtype: tuple
+    """
+    # Determine extension on input file
+    base, ext1 = os.path.splitext(url)
+    _, ext2 = os.path.splitext(base)
+
+    # Produce or download BAM file
```
fnothaft: I'm a real pedant about this, but can you add comments describing the control flow for each fork of the if/else statement? Also, I might note that the logic in the `.bam` extension fork of the if/else isn't necessarily what an end user might expect. E.g., if you provide a BAM and set `config.run_bwa` to true, do you want to just download the BAM file? Possibly, but you might also want to realign it --> bwakit supports that on the command line. We should be more explicit.

fnothaft: Also, minor nit, but ideally you'd support .fq.gz/.fastq.gz/etc and variants with different case (e.g., .FQ)

src/toil_scripts/gatk_germline/germline.py
```diff
@@ -1,304 +1,500 @@
...
+
+    rm_secondary = job.wrapJobFn(run_samtools_view,
+                                 get_bam.rv(),
+                                 flag='0x800',
+                                 ncores=config.cores,
+                                 disk=config.file_size, cores=config.cores)
+
+    sort_bam_disk = PromisedRequirement(lambda x: 3 * x.size + human2bytes('10G'),
+                                        rm_secondary.rv())
+    sort_bam = job.wrapJobFn(run_samtools_sort,
```
fnothaft: You may want to make sorting optional, since someone may provide a sorted BAM as input.

src/toil_scripts/gatk_germline/germline.py
```diff
@@ -1,304 +1,500 @@
...
+                             rm_secondary.rv(),
+                             ncores=config.cores,
+                             cores=config.cores,
+                             disk=sort_bam_disk)
+
+    index_bam_disk = PromisedRequirement(lambda x: int(1.5 * x.size), sort_bam.rv())
```
fnothaft: Probably just need 1x size.

src/toil_scripts/gatk_germline/germline.py
```diff
@@ -1,304 +1,500 @@
...
+    job.fileStore.logToMaster("Aligning sample: {}".format(config.uuid))
+    cores = multiprocessing.cpu_count()
+    # Download fastq files
+    fq1 = job.addChildJobFn(download_url_job,
+                            url,
+                            name='toil.1.fq',
+                            synapse_login=config.synapse_login,
+                            s3_key_path=config.ssec,
+                            disk=config.file_size)
+    # Assumes second fastq url is identical
```
fnothaft: This is going to give someone the business in the future. Can we validate the URL prior to attempting the download? If it doesn't exist, we should throw a meaningful error.

src/toil_scripts/gatk_germline/germline.py
```diff
@@ -1,304 +1,500 @@
...
+    bwa_config.r2 = fq2.rv()
+    # bwa_alignment uses a different naming convention
+    bwa_config.ref = config.genome_fasta
+    bwa_config.fai = config.genome_fai
+    bwakit_disk = PromisedRequirement(lambda x, y: 3 * (x.size + y.size) + human2bytes('8G'),
+                                      fq1.rv(), fq2.rv())
+    return job.addFollowOnJobFn(run_bwakit,
+                                bwa_config,
+                                cores,
+                                sort=False,
```
fnothaft: OOC, what's the advantage to disabling sorting here, and sorting above?

fnothaft:
Thanks @jpfeil! I just made a targeted pass over this. I'll make more of a pass tomorrow. Would it be possible for you to rebase this on #335? It looks like most of the changes are consistent between the two. Ideally, this would be one commit on top of #335. It is also OK with me if you squash #335 down. This would just make reviewing easier.

src/toil_scripts/gatk_germline/germline.py
jpfeil: I thought it would be simpler to have one point in the pipeline where all samples get sorted.

src/toil_scripts/gatk_germline/germline.py
```diff
@@ -1,304 +1,500 @@
...
+    sort_bam_disk = PromisedRequirement(lambda x: 3 * x.size + human2bytes('10G'),
+                                        rm_secondary.rv())
+    sort_bam = job.wrapJobFn(run_samtools_sort,
+                             rm_secondary.rv(),
+                             ncores=config.cores,
+                             cores=config.cores,
+                             disk=sort_bam_disk)
+
+    index_bam_disk = PromisedRequirement(lambda x: int(1.5 * x.size), sort_bam.rv())
+    index_bam = job.wrapJobFn(run_samtools_index, sort_bam.rv(), disk=index_bam_disk)
```
fnothaft: Also, I think `samtools sort` writes an index file. You may be able to remove this.

src/toil_scripts/gatk_germline/germline.py
fnothaft: Ah, that's a reasonable decision! Thanks for explaining such. Can you document that inline?

src/toil_scripts/adam_gatk_pipeline/align_and_call.py
```diff
@@ -124,8 +124,7 @@
 # import job steps from other toil pipelines
 from toil_scripts.adam_pipeline.adam_preprocessing import * #static_adam_preprocessing_dag
 from toil_scripts.bwa_alignment.bwa_alignment import * #download_shared_files
-from toil_scripts.gatk_germline.germline import * #batch_start
-from toil_scripts.gatk_processing.gatk_preprocessing import * #download_gatk_files
+from toil_scripts.gatk_germline.germline import * #download_and_run
```
jvivian: I'm guessing @fnothaft has a good reason, but star imports are generally a no-no: https://pythonconquerstheuniverse.wordpress.com/2011/03/28/why-import-star-is-a-bad-idea/

fnothaft: This is due to a really nasty serialization bug in Toil that I don't think has ever been fixed...

src/toil_scripts/exome_variant_pipeline/exome_variant_pipeline.py
```diff
@@ -20,7 +20,7 @@
 from toil_scripts.tools.mutation_callers import run_muse
 from toil_scripts.tools.mutation_callers import run_mutect
 from toil_scripts.tools.mutation_callers import run_pindel
-from toil_scripts.tools.preprocessing import run_gatk_preprocessing
+from toil_scripts.tools.preprocessing import run_germline_preprocessing
```
jvivian: Semantics, but since this is being used in the exome pipeline, why `run_germline_preprocessing` over `run_gatk_preprocessing` ?

fnothaft: +1

src/toil_scripts/gatk_germline/germline.py
```diff
@@ -1,688 +1,669 @@
...
+
+    :param JobFunctionWrappingJob job: Toil Job instance
+    :param str uuid: Unique identifier for the sample
+    :param str url: URL to BAM file
+    :param Namespace config: Configuration options for pipeline
+    :param str rg_line: RG line for BWA alignment. Default is None.
+    :return: None
+    """
+    get_shared_files = job.wrapJobFn(download_shared_files, config).encapsulate()
+    samples = [(uuid, url, url2, rg_line)]
```
jvivian: Why nest a tuple inside a list?

src/toil_scripts/gatk_germline/germline.py
```diff
@@ -1,688 +1,669 @@
...
+    run_samtools_index, run_samtools_view, run_picard_create_sequence_dictionary
+from toil_scripts.tools.variant_annotation import run_oncotator
+
+
+def download_and_run_sample(job, uuid, url, config, url2=None, rg_line=None):
+    """
+    Downloads reference files and runs the GATK best practices germline pipeline for one sample.
+    See the germline config file for necessary parameters.
+
+    :param JobFunctionWrappingJob job: Toil Job instance
```
jvivian: Might add that this is passed to every job automatically (other pipelines use this comment at Hannes's suggestion)

jvivian:
I'm guilty of this as well, but reviewing one commit, associated with one particular change or feature, is a lot easier than line commenting through the file diffs when there are so many changes.

src/toil_scripts/gatk_germline/germline.py
```diff
@@ -1,688 +1,669 @@
...
+    vqsr_name = '{}.raw{}.gvcf'.format(uuid, config.suffix)
+    output_gvcf = job.wrapJobFn(upload_or_move_job,
+                                vqsr_name,
+                                haplotype_caller.rv(),
+                                output_dir)
+    haplotype_caller.addChild(output_gvcf)
+    return haplotype_caller.rv()
+
+
+def download_shared_files(job, config):
```
jvivian: I generally place job functions in a hierarchical order from top to bottom — e.g. since this is called by `startToil` it would be at the top, follow by `reference_preprocessing`

src/toil_scripts/gatk_germline/germline.py
jpfeil: The pipeline is designed to run a set of samples, but in this function we are only running one sample.

jvivian:
@jpfeil — Can you ping me for review once tests pass / rebase onto master?

jpfeil:
This PR is not needed because Oncotator has been added in #335

PR #421, opened 2016-08-22
Remove redundant passing of cores (resolves #374)

### LINKED ISSUE #374: Remove redundant passing of "cores"
Job functions can access the number of cores they were provisioned with `job.cores`, which means cores shouldn't be passed around as an argument to job functions redundantly.

jvivian:
Remove cores from SplAdder pipeline
Remove cores from s3am_upload_job
Remove cores from RNA-seq and tools
Remove cores from exome pipeline and tools
Remove cores from BWA pipeline and tools


fnothaft:
LGTM!

PR #440, opened 2016-09-08
Add dynamic disk requirements (resolves #216)

### LINKED ISSUE #216: Add dynamic disk resource allocation for rnaseq_cgl

jvivian:
Doesn't resolve 2 resource requirements:  

- Initial size of sample being downloaded (for example, from an http URL)
- STAR memory requirement based off size of STAR index which is downloaded within the STAR job function.


jvivian:
@jpfeil — I'll probably bug you tomorrow about help debugging this.

PR #442, opened 2016-09-08
Create output-dir if it does not exist (resolves #441)

### LINKED ISSUE #441: toil-bwa should create output dir if not exists
Currently the config states
```
# Required: Output location of sample. Can be full path to a directory or an s3:// URL
# Warning: S3 buckets must exist prior to upload or it will fail.
output-dir: 
```
Note that it does not say anything about a local output directory. Other toil-scripts create it if it does not exist, but toil-bwa runs for 3 hours (12 nodes) and *then* fails with `IOError: [Errno 2] No such file or directory: '/tmp/bwa/out1/local.bam'`
(I don't know why it lists the file, not the directory)

Request: Either have the code create the directory or change the config.

jvivian:


jvivian:

PR #449, opened 2016-09-13
README now links to dockerized pipeline README

alex-hancock:


hannes-ucsc:
Doc change only. Merging now.

PR #456, opened 2016-09-14
Readme updates

alex-hancock:
Added docker links to exome and bwa pipeline READMEs.


hannes-ucsc:
Added docker links to exome and bwa pipeline READMEs

PR #459, opened 2016-09-15
Stop virtualenv from upgrading setuptools (resolves #458)

### LINKED ISSUE #458: Jenkins builds are not deploying to pypi
Same issue as in https://github.com/BD2KGenomics/toil-lib/issues/22

cket:


cket:
@hannes-ucsc

PR #463, opened 2016-09-16
Bump toil-lib version to 1.0.4a1.dev55 (resolves #462)

cket:

PR #465, opened 2016-09-16
Refer to proper version of toil-lib

cket:
The proper version was hidden on Pypi. @hannes-ucsc fixed that and this change updates the wrong version included in #463


cket:
@hannes-ucsc

